PR #6, opened 2023-09-13
Adding the Swap feature 

spalmer25:
This PR enables the Ledger `swap` function:
- [x] Add ENABLE_SWAP tag
- [x] Provide the 4 required functions:
  * [x] The `swap_handle_check_address` function which checks an address.
  * [x] The `swap_handle_get_printable_amount` function, which gives the print format for an amount.
  * [x] The `swap_copy_transaction_parameters` function, which backs up transaction parameters.
  * [x] The `swap_finalize_exchange_sign_transaction` function which finalizes the exchange.
- [x] During a `swap`, check that the transaction matches the backed-up parameters.

I made some choices in the implementation:
 - In `check_address`: the `derivation_type` is set to `ED25519` (since no parameter defines it).
 - In `check_address`/`copy_transaction_parameters`: the `extra_id` must be non-null and empty.
 - In the `get_printable_amount` function: the `ticker` and the number of decimals given by the caller are those used for print. Should we instead force the display with `tz` and 6 decimals?
 - In `copy_transaction_parameters`/`check_validity`: the `ticker` and the number of decimals are not taken into account, the `amount`/`fee` values are just compared without decimals.
 
 The feature tests are on an `LedgerHQ/app-exchange` fork: https://github.com/spalmer25/app-exchange/pull/1


spalmer25:
Peer on GitLab:
https://gitlab.com/nomadic-labs/tezos-ledger-app-revamp/-/merge_requests/139

emturner (commented):
overall LGTM, will test manually against ledger live

app/src/handle_swap.c
```diff
@@ -0,0 +1,85 @@
...
+
+  if (params->extra_id_to_check[0] != '\0') {
+    PRINTF("[ERROR] Extra id is not empty: %s\n", params->extra_id_to_check);
+    goto error;
+  }
+
+  char address[TZ_CAPTURE_BUFFER_SIZE] = {0};
+
+  // Always tz1 ?
+  derivation_type_t derivation_type = DERIVATION_TYPE_ED25519;
```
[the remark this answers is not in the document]
spalmer25: Perfect, I'll stick to that for this PR.

app/src/keys.c
```diff
@@ -134,6 +134,33 @@ public_key_hash(uint8_t *hash_out, size_t hash_out_size,
...
+    TZ_ASSERT_NOTNULL(buffer);
+    TZ_CHECK(generate_public_key(&pubkey, derivation_type, bip32_path));
+    TZ_CHECK(public_key_hash(hash + 1, 20, NULL, derivation_type, &pubkey));
+    // clang-format off
+    switch (derivation_type) {
+    case DERIVATION_TYPE_SECP256K1: hash[0] = 1; break;
+    case DERIVATION_TYPE_SECP256R1: hash[0] = 2; break;
+    case DERIVATION_TYPE_ED25519:
+    case DERIVATION_TYPE_BIP32_ED25519: hash[0] = 0; break;
+    default: CX_CHECK(EXC_WRONG_PARAM); break;
```
elric1, line 154: CX_CHECK should be TZ_FAIL here.

app/src/keys.h
```diff
@@ -67,3 +67,5 @@ check_derivation_type(derivation_type_t code)
 {
     return (code >= DERIVATION_TYPE_ED25519 && code < DERIVATION_TYPE_MAX);
 }
+
+void generate_pkh(derivation_type_t, const bip32_path_t *, char *, size_t);
```
elric1, line 71: It would look better to put this prototype with the other prototypes after `generate_public_key()`.

app/src/keys.c
```diff
@@ -134,6 +134,33 @@ public_key_hash(uint8_t *hash_out, size_t hash_out_size,
     TZ_POSTAMBLE;
 }
 
+void
```
elric1, line 137: The function above (`public_key_hash()`) should be static as we've removed its prototype and it therefore should not be accessible outside this C file.

elric1 (changes requested):

app/src/parser/num_parser.c
```diff
@@ -66,3 +67,67 @@ tz_parser_result tz_parse_nat_step(tz_num_parser_buffer *buffers,
...
+  out[offset--] = '\0';
+
+  // drop trailing non significant zeroes
+  while (out[offset] == '0') out[offset--] = '\0';
+  if (out[offset] == '.') out[offset--] = '\0';
+
+  return true;
+}
+
+bool tz_print_uint64(uint64_t value, char *out, int out_len, int nb_decimals) {
```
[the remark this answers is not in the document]
elric1: We should not use "print" in these function names, because they do not print.
tz_print_mutez() can be replaced by format_fpu64() from lib_standard_app in ledger-secure-sdk, I think.
If you are worried about building it outside for, say, tests, then use the approach in app/src/parser/compat.h and define:

```
bool
tz_format_mutez(uint64_t value, char *out, size_t outlen)
{
#ifdef ACTUALLY_ON_LEDGER
    return format_fpu64(out, outlen, value, 6);
#else
    size_t len = snprintf(out, outlen, "%u.%06u", value / 1000000, value % 1000000);
    return len < outlen;
#endif
}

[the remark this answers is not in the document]
Comment (elric1) [reply app/src/parser/num_parser.c]: I'm going to add the above comment to my review, if I can.

File: app/src/parser/num_parser.c [hunk 6]
```diff
@@ -75,3 +76,107 @@ tz_parse_nat_step(tz_num_parser_buffer *buffers, tz_num_parser_regs *regs,
...
+    while (power != 1) {
+        power /= 10;
+        buff[offset++] = value / power + '0';
+        value %= power;
+    }
+
+    buff[offset] = '\0';
+
+    return tz_adjust_decimal(buff, offset, out, out_len, nb_decimals);
+}
```
elric1, line 154: This duplicates functionality that is already defined in `ledger-secure-sdk/lib_standard_app/format.c`. This function should be able to be rewritten something like this:
```
bool
tz_print_mutez2(uint64_t value, char *out, int out_len, int nb_decimals)
{
#if ACTUALLY_ON_LEDGER
    format_fpu64(out, out_len, value, nb_decimals);
#else
    size_t divisor = 1;
    size_t i;
    size_t ret;

    for (i=0; i < nb_decimals; i++)
        divisor *= 10;

    ret = snprintf(out, out_len, "%u.%0*u",
                   value / divisor,  nb_decimals, value % divisor);
    return ret < out_len;
#endif
}

```

elric1 (changes requested):
I noticed another quick one. This should simplify things quite a lot.

app/src/parser/num_parser.c
elric1: `ACTUALLY_ON_LEDGER` can be found in `app/src/parser/compat.h` which is how you decide if you are inside the unit tests.
My function assumes that if you are not on a ledger, then you are on an LP64 box with a functioning `snprintf(3)` in libc.

app/src/parser/num_parser.c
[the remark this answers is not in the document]
spalmer25: Perfect!

app/src/parser/num_parser.c
spalmer25: This function is only useful in swap, so it would be better not to have this function in parser at all and to use the functions from `ledger-secure-sdk/lib_standard_app/format.c` directly.

elric1 (approved):

PR #9, opened 2023-09-18
app: store settings in NVM

emturner:
required to persist settings (currently just blindsiging toggle) between user-runs of the app.

# Testing

The following works with the inclusion of this PR:
- Run the app on a physical device (e.g. nanos)
- Open the app
- enable blindsigning
- exit the app
- Re-open the app. Blindsigning should still be enabled

To build & run the app on the nanos, connect & unlock the nanos, and then run:
```
docker run --rm -ti --user "$(id -u):$(id -g)" --privileged -v "/dev/bus/usb:/dev/bus/usb" -v "$(realpath .):/app" ghcr.io/ledgerhq/ledger-app-builder/ledger-app-dev-tools:latest

cd app
make load
```
and follow the prompts on the nanos

### DIFF
--- a/app/src/globals.c
+++ b/app/src/globals.c
@@ -1,6 +1,7 @@
 /* Tezos Ledger application - Global application state
 
    Copyright 2023 Nomadic Labs <contact@nomadic-labs.com>
+   Copyright 2023 Trilitech <contact@trili.tech>
 
    With code excerpts from:
     - Legacy Tezos app, Copyright 2019 Obsidian Systems
@@ -25,9 +26,22 @@
 
 globals_t global;
 
+const settings_t N_settings_real;
+
 void init_globals(void) {
     memset(&global, 0, sizeof(global));
     memset(G_io_seproxyhal_spi_buffer, 0, sizeof(G_io_seproxyhal_spi_buffer));
     memset(&G_ux, 0, sizeof(G_ux));
     memset(&G_ux_params, 0, sizeof(G_ux_params));
+}
+
+void toggle_blindsigning(void) {
+    settings_t tmp;
+    memcpy(&tmp,
+           (void*) &N_settings,
+           sizeof(tmp));
+    tmp.blindsigning = !N_settings.blindsigning;
+    nvm_write((void*) &N_settings,
+              (void*) &tmp,
+              sizeof(N_settings));
 }
--- a/app/src/ui_home_nbgl.c
+++ b/app/src/ui_home_nbgl.c
@@ -52,7 +52,7 @@ static bool navigation_cb_wallet(__attribute__((unused))uint8_t page,
   case 1:
     switches[0] =
       (nbgl_layoutSwitch_t)
-      {.initState = global.settings.blindsigning ? ON_STATE : OFF_STATE,
+      {.initState = N_settings.blindsigning ? ON_STATE : OFF_STATE,
        .text = "Blind signing",
        .subText = "Enable blindsigning",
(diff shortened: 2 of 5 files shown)


app/src/ui_settings.c
```diff
@@ -14,10 +31,14 @@ static void
 cb(tz_ui_cb_type_t type)
 {
   FUNC_ENTER(("type=%u\n", type));
+  bool new_blindsigning;
 
   switch (type) {
   case BLIND_SIGNING:
-    global.settings.blindsigning = !global.settings.blindsigning;
+    new_blindsigning = !N_settings.blindsigning;
```
elric1, line 38: I think that we should rewrite the entire struct here. I don't feel comfortable that we can take an & of a bool and write to it without having implementation defined or undefined behaviour.

elric1: So maybe:
```
    nvm_write(&N_settings, new_settings, sizeof(N_settings));
```
I think that the `sizeof()` should be of the first argument as that is what determines the size if there is a mismatch.

app/src/ui_home_nbgl.c
```diff
@@ -71,10 +71,15 @@ static bool navigation_cb_wallet(__attribute__((unused))uint8_t page,
 
 static void controls_callback(int token,
                               __attribute__((unused))uint8_t index) {
+  bool new_blindsigning;
   switch (token) {
   case BLIND_SIGNING_TOKEN:
-    global.settings.blindsigning = !global.settings.blindsigning;
-    if (!global.settings.blindsigning)
+    new_blindsigning = !N_settings.blindsigning;
```
elric1, line 77: As below:
```
    settings_t tmp;
    memcpy(&tmp, &N_settings, sizeof(tmp));
    tmp.blindsigning = !N_settings.blindsigning;
    nvm_write(&N_settings, &tmp, sizeof(N_settings);
```
or something like this.
Given that it repeats below, maybe it would do well to make this code block a function in globals.c?

elric1 (changes requested):

app/src/ui_home_nbgl.c
elric1: That said, maybe it's hard to do a single function. `flip_blindsigning()` would make it feel like an OO sort of approach with accessor methods...

emturner: added `toggle_blindsigning`

app/src/ui_settings.c
emturner: added `toggle_blindsigning`

app/src/globals.c
```diff
@@ -25,9 +26,22 @@
...
     memset(G_io_seproxyhal_spi_buffer, 0, sizeof(G_io_seproxyhal_spi_buffer));
     memset(&G_ux, 0, sizeof(G_ux));
     memset(&G_ux_params, 0, sizeof(G_ux_params));
 }
+
+void toggle_blindsigning(void) {
+    settings_t tmp;
+    memcpy((settings_t*) &tmp,
+           (settings_t*) &N_settings,
+           sizeof(tmp));
```
elric1, line 42: You don't need to cast here, because `memcpy(3)` takes `void *` and `const void *` and so it should implicitly cast your type `settings_t *` to `void *` without a warning.

elric1 (changes requested):

app/src/globals.c
emturner: volatile causes some issues here, so the cast on `N_settings` is needed

elric1 (approved):

PR #12, opened 2023-09-20
Just a bit of code cleanup on the way to clang-format

elric1:


app/src/parser/parser_state.c
```diff
@@ -17,20 +17,24 @@
 
 #include "parser_state.h"
 
+#define TZ_LABEL(_x)  case TZ_ ## _x: return #_x
+#define BLO_LABEL(_x) case TZ_BLO_ ## _x: return #_x
```
emturner, line 21: how come spaces are okay here? Intuitively might expect it to have to be `TZ_##_x`?

emturner: ignore me, misread the docs

emturner (approved):

PR #15, opened 2023-09-21
Add the code quality github workflow from app-boilerplate

elric1:


.github/workflows/codeql_checks.yml
```diff
@@ -0,0 +1,45 @@
...
+    paths-ignore:
+      - '.github/workflows/*.yml'
+      - 'tests/*'
+
+jobs:
+  analyse:
+    name: Analyse
+    strategy:
+      matrix:
+        sdk: [ "$NANOS_SDK", "$NANOX_SDK", "$NANOSP_SDK" ]
```
emturner, line 21: does this support the stax sdk too?

.github/workflows/codeql_checks.yml
```diff
@@ -0,0 +1,45 @@
...
+          queries: security-and-quality
+
+      # CodeQL will create the database during the compilation
+      - name: Build
+        run: |
+          make BOLOS_SDK=${{ matrix.sdk }} -C app
+          make BOLOS_SDK=${{ matrix.sdk }} -C app DEBUG=1
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v2
```
emturner, line 45: could this be integrated into the `build_and_test` workflow potentially? with this we're now building the app twice

elric1: well... that's an interesting point. I am just copying what the app-boilerplate is doing and so if we want to deviate, I think that it would be a good idea to chat with them.

.github/workflows/codeql_checks.yml
elric1: I would assume so. Their template didn't have it, I have to admit that I didn't even check.

.github/workflows/codeql_checks.yml
emturner: let's merge for now, have raised #17

emturner (approved):

PR #18, opened 2023-09-22
STAX: clearsigning flow

emturner:
Clearsigning for STAX is slightly more complex, due to having to support the 'skip' flow, that allows a user to go directly to the signing page.


app/src/ui_stream.h
```diff
@@ -40,8 +40,14 @@
...
 #define TZ_UI_STREAM_HISTORY_SCREENS 8
 #define TZ_UI_STREAM_TITLE_WIDTH     TZ_SCREEN_WITDH_BETWEEN_ICONS_BOLD_11PX
-#define TZ_UI_STREAM_CONTENTS_WIDTH  TZ_SCREEN_WITDH_FULL_REGULAR_11PX
-#define TZ_UI_STREAM_CONTENTS_LINES  (TZ_SCREEN_LINES_11PX - 1)
+
+#ifdef HAVE_BAGL
+#define TZ_UI_STREAM_CONTENTS_WIDTH TZ_SCREEN_WITDH_FULL_REGULAR_11PX
+#define TZ_UI_STREAM_CONTENTS_LINES (TZ_SCREEN_LINES_11PX - 1)
+#elif HAVE_NBGL
+#define TZ_UI_STREAM_CONTENTS_WIDTH 60
```
spalmer25, line 52: Why 60?
I ran the sign_execute_outbox_messages test manually with 120 and it seemed to work.
Is this a value based on the maximum size characters?

tests/integration/stax/test_sign_execute_outbox_messages.py
```diff
@@ -0,0 +1,64 @@
...
+    app.review.tap()
+    app.assert_screen("operation_commitment_src12UJ...LY8")
+
+    app.review.tap()
+    app.expect_apdu_return("9000")
+
+    app.send_apdu("800f82004f63313162343430616334643334353564656462653465653064653135613861663632306434633836323437643964313332646531626236646132336435666639643864666664613232626139613834")
+
+    for i in range(7):
+        app.review.tap()
```
spalmer25, line 54: Here, you `tap` two times before changing screens.
Once before sending the new apdu and once after.

emturner: Hi, good spot! Yes, this is the case. 

This falls under one of those 'MVP' issues - the fix would take a little bit of work to do. I can raise an issue if we're happy to tackle separately?

app/src/ui_stream.h
emturner: A rough guesstimate at being able to fit things on a single screen where needed (e.g. hashes), vs not over-filling the screen.

I am planning on raising a follow up issue to tackle this properly, in the same style as the nano devices - but might be cleaner as a follow up.

wdyt?

spalmer25: Indeed, it is not a priority to do it here.
Further discussion on this subject will not be excessive.

tests/integration/stax/test_sign_execute_outbox_messages.py
spalmer25: Absolutely!

tests/integration/stax/utils.py
```diff
@@ -66,6 +68,12 @@ def expect_apdu_return(self, expected):
         expected = bytes.fromhex(expected)
         assert response == expected, f"Expected {expected}, received {response}"
 
+    def expect_apdu_failure(self, code):
+        """Expect 9001 failure"""
```
spalmer25, line 72:
```suggestion
        """Expect code failure"""
```

tests/integration/stax/test_reject_review_operation.py
```diff
@@ -0,0 +1,42 @@
...
+
+    app.assert_screen("review_request_sign_operation")
+
+    app.review_reject_signing()
+
+    app.review.tap()
+    app.assert_screen(SCREEN_HOME_DEFAULT)
+
+    app.send_apdu("800f81005e0300000000000000000000000000000000000000000000000000000000000000006c016e8874874d31c3fbd636e924d5a036a43ec8faa7d0860308362d80d30e01000000000000000000000000000000000000000000ff02000000020316");
+    app.expect_apdu_failure("9001")
```
spalmer25, line 39: Why is it the `EXC_UNEXPECTED_STATE` error?
Shouldn't it be the `EXC_REJECT` error (=6985)?

emturner: You're right, it _should_ be `EXC_REJECT`. I think there's something a bit weird going on when you reject after the first apdu - the app isn't really set up to handle that well.

I'll have a look

tests/integration/stax/utils.py
emturner: fixed in 9460dd394339ffb37cc3581442e73887744753c8

tests/integration/stax/test_reject_review_operation.py
emturner: fixed in 9460dd394339ffb37cc3581442e73887744753c8

There was a simple explanation - because we were sending the second apdu on the _home_ screen, we correctly got an unexpected state error, since the app has been reset and we're sending the 'second' apdu packet.

Whereas if we send the second apdu immediately, it gives the chance for the app to respond with the correct code. (This is what happens in actuality)

app/src/apdu_sign.c
```diff
@@ -165,6 +166,13 @@ refill()
 
         tz_parser_flush_up_to(st, global.line_buf, TZ_UI_STREAM_CONTENTS_SIZE,
                               wrote);
+
+        // Do as much parsing as we can in one go when skipping.
+        // TODO: perhaps when skipping we could only skip the current
+        //       operation, rather than the whole rest-of-batch
+        if (global.apdu.sign.u.clear.skip_to_sign)
+            goto skip;
```
elric1, line 174: We will be able to blow past all of the history here with no user input. Are we sure that this isn't a problem? What I mean is that we could very well get into a situation where s->current is way behind and refers to what is effectively a random page in the mix.

elric1 (commented):
Quick comment, I'll keep reviewing, though.

app/src/apdu_sign.c
emturner: This shouldn't be a problem for now - since this workflow is only available on `stax` - and there's no way for the user to go back to the previous screen in the current workflow

elric1: I wonder if it might be better to just export `change_screen_{left,right}()` and `change_screen_left()` here with some consideration. This would make the skip functionality essentially the same as paging through the output without displaying it---and that would require less analysis to convince ourselves that we got correct.

elric1: (This would introduce the requirement that no single `tz_ui_stream_push()` could add more than a single screen, which I think that we have at the moment, but it's not documented to be a requirement.)

PR #22, opened 2023-09-25
build: bump sdk/speculos versions

emturner:
includes:
- OCR fix for missing 'S' / conversion of 'I' to 'l'

introduces:
- weird OCR bug on nanox around 'Title' not being picked up


tests/integration/check_section_text.py
```diff
@@ -27,18 +27,19 @@ class Screen:
...
+      content = content.lstrip('\n')
+      print(f"l: {l}, c: {content[:len(l)]}")
       if not content.startswith(l):
         return False
       content = content.removeprefix(l)
       content = content.lstrip('\n')
 
-    return len(self.text) == content_lines or len(content) == 0
+    # blank content lines may not show up
+    return len(self.text) <= content_lines or len(content) == 0
```
spalmer25, line 42: I could be wrong, but this `len(self.text) <= content_lines` condition still seems `true` to me, or at least true to what I expected.
If we receive `text` from the application, we can be pretty sure that it will never have more lines than `content_lines`.

The old `len(self.text) == content_lines` condition was only there to ensure that all lines were displayed if the `text` was not full.
This condition was useful because sometimes (the flacky tests), we only received the start of the text (the title and one or two lines when there should have been more) this matched but only removed the text received from the `content` (and not all the lines it should have received) and so the test failed on the next screen.

But if the empty lines are not displayed, then this fix is no longer relevant.

In any case, I think the `len(self.text) <= content_lines` condition could just be `true`.
And we'll have to see if the new version of speculos fixes the old problem.

emturner: I think you're completely right - will remove

spalmer25 (approved):

PR #29, opened 2023-09-27
Random cleanups

elric1:
[Fix a couple of issues with tz_string_to_mutez()](https://github.com/trilitech/ledger-app-tezos-wallet/commit/c5c7573e6d499ac8c5db75867321c61989b4fa98): this one is actually important. It won't scan-build without it, because we might deref a NULL ptr.

[Makefile: add a scan-build target](https://github.com/trilitech/ledger-app-tezos-wallet/commit/5f9258e5be52cb445c39f315e7cf931c711fddae): this is useful, because we will be adding this to the CI, not via the Makefile, but this makes it easier to run it locally.


Makefile
```diff
@@ -40,6 +40,17 @@ docker_images:	docker_speculos			\
 		docker_ledger_app_ocaml		\
 		docker_ledger_app_integration_tests
 
+scan-build-%:
+	if [ -d "$(BOLOS_SDK)" ]; then                                    \
+	    make -C app scan-build;                                       \
+	else                                                              \
+	    $(DOCKER_RUN_APP_BUILDER) bash -c                             \
+		"BOLOS_SDK=/opt/nanos-secure-sdk make -C app scan-build"; \
```
spalmer25, line 48: `opt/nanos-secure-sdk` is used even for `scan-build-nanosp`, `scan-build-nanox` and `scan-build-stax`.
Shouldn't `BOLOS_SDK` correspond to the device?

elric1: Yes, I forgot that because it was a bit late last night. I've added it in and also simplified the `app_%{,_dbg}.tgz` logic a little bit. Just a little...

app/src/apdu_sign.c
```diff
@@ -150,63 +150,70 @@ send_continue(void)
...
+               tz_parser_result_name(st->errno));
+        // clang-format off
+	switch (st->errno) {
+	case TZ_BLO_IM_FULL: TZ_CHECK(refill_blo_im_full()); break;
+	case TZ_BLO_FEED_ME: TZ_CHECK(send_continue());      break;
+	case TZ_BLO_DONE:    TZ_CHECK(refill_blo_done());    break;
+	default:             TZ_CHECK(refill_error());       break;
+	}
+        // clang-format on
+    } while (st->errno == TZ_BLO_IM_FULL && skip_to_sign);
```
spalmer25, line 216: This refactoring does not produce the same behaviour as before:
If `st->errno == TZ_BLO_DONE` and `st->regs.oofs != 0` and `global.apdu.sign.u.clear.skip_to_sign` then it should loop back, but in this new version it doesn't.

elric1: I am eliminating the loops, though. So, what we do is:
```
    if (st->regs.oofs != 0)
        refill_blo_im_full();
```
in `refill_blo_im_full()`.

spalmer25: Indeed, if `st->errno == TZ_BLO_DONE` and `st->regs.oofs != 0` then `refill_blo_im_full()` is called.
But once this function has terminated the initial loop `} while (st->errno == TZ_BLO_IM_FULL && skip_to_sign);`
stops since `st->errno == TZ_BLO_DONE` and this even if `skip_to_sign == true` whereas in the old version another loop cycle would have been started  if `skip_to_sign == true`.

spalmer25 (approved):

PR #32, opened 2023-09-29
check_section_text.py: our previous fix wasn't enough

elric1:


spalmer25 (approved):

tests/integration/check_section_text.py
```diff
@@ -29,15 +29,21 @@ class Screen:
...
-      content = content.lstrip('\n')
-      if len(l) == 0 or not content.startswith(l):
+      if not content.startswith(l):
         return False
       content = content.removeprefix(l)
+      content = content.lstrip('\n')
+      consumed += len(l)
 
-    return True
+    return consumed > min_chars or len(content) == 0
```
spalmer25, line 46: If I understand correctly, `consumed > min_chars` is a condition to ensure that if we haven't parsed the entire `content` yet, the screen should be full enough.
Why not just make sure that the number of lines displayed matches `content_lines` (the maximum number of lines) instead of relying on `min_char` (a heuristic of the number of characters displayed on those lines)?

elric1: Because the last page might be short.

PR #34, opened 2023-09-29
apdu_pubkey.c: don't check if we are locked

elric1:
The Ledger developers informed us that we do not need to do this.

We were informed that this is "probably" an unnecessary check. We should make sure that we agree before we put this in.


ajinkyaraj-23:
After reviewing BOLOS documentation, I can confirm that we don't need this check. Please refer to to this [link](https://developers.ledger.com/docs/embedded-app/bolos-features/#passphrases-in-bolos).

emturner (approved):

PR #40, opened 2023-10-02
Fix the disappearing public key on stax

### LINKED ISSUE #38: Stax: PROMPT_PUBLIC_KEY address disappears
On the `PROMPT_PUBLIC_KEY` request, on the confirmation screen, the address disappears after coming back from the QR screen.

# Goal
Screenshot from 2023-10-02 10-03-12

# Current behaviour
Screenshot from 2023-10-02 09-22-52

spalmer25:
fixes #38


tests/integration/stax/test_provide_pk.py
```diff
@@ -26,6 +26,13 @@
     app.provide_pk.tap()
     app.assert_screen("screen_show_address_tz1_zebra")
 
+    # tap on "Show as QR"
+    app.provide_pk.client.finger_touch(200, 275)
```
emturner, line 30: If possible, it would be great to use a pre-defined `position` - does `CENTER` work here? 

If not, it would be good to have the position explicitly defined (maybe in utils?)

spalmer25: Unfortunately, the `Show as QR` button is positioned just after the public key, so it's not quite in the middle and, more importantly, depends on the size of the public key.
I'll define explicitly the position in its own test file.

emturner (approved):

PR #41, opened 2023-10-03
Stax: add a spinner on skip

spalmer25:
Currently, after a skip during a signing request on stax, the user is able to sign even though all the apdus have not yet been received.
To avoid this, this MR adds a spinner until all the apdus have been received.

### DIFF
--- a/app/src/ui_stream_nbgl.c
+++ b/app/src/ui_stream_nbgl.c
@@ -1,6 +1,7 @@
 /* Tezos Ledger application - Generic stream display
 
    Copyright 2023 Nomadic Labs <contact@nomadic-labs.com>
+   Copyright 2023 Functori <contact@functori.com>
    Copyright 2023 TriliTech <contact@trili.tech>
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


tests/integration/stax/test_sign_execute_outbox_messages_skip.py
```diff
@@ -35,6 +35,8 @@
 
     app.review_skip_to_signing()
 
+    app.assert_screen("skipping")
```
emturner, line 38: I think there's a bit of a race condition here potentially. Because the ticker isn't paused, the loading screen will spin round.

Instead, it might be a good idea to pause the ticker and then assert that the loading screen is displayed?

emturner, line 38:
```suggestion
    app.assert_screen("loading_operation")
```

app/src/ui_stream_nbgl.c
```diff
@@ -195,7 +195,15 @@ tz_ui_nav_cb(uint8_t page, nbgl_pageContent_t *content)
...
     }
 
-    if (!s->full && !global.apdu.sign.u.clear.skip_to_sign) {
+    if (s->full) {
+        content->type                        = INFO_LONG_PRESS;
+        content->infoLongPress.icon          = &C_tezos;
+        content->infoLongPress.text          = "Sign";
+        content->infoLongPress.longPressText = "Sign";
+    } else if (page == LAST_PAGE_FOR_REVIEW) {
+        nbgl_useCaseSpinner("Skipping");
```
emturner, line 204:
```suggestion
        nbgl_useCaseSpinner("Loading operation");
```
for consistency with blindsigning loading screen?

emturner:
I wonder if we should have a `is last pkt` check for signing, the way we do for `reject`?

spalmer25:
> I wonder if we should have a `is last pkt` check for signing, the way we do for `reject`?

In fact, this is already the case

emturner (approved):

PR #46, opened 2023-10-05
Re-enable nanox generated tests

### LINKED ISSUE #43: Nanox: re-enable generated tests
Disabled in #22.

The OCR system was completely re-done in the SDK version used in #22 - rather than using OCR, instead the actual bitmaps are extracted from the elf binaries and are used to get info off the screen. (See https://github.com/LedgerHQ/speculos/pull/418 for context)

We will likely have to talk to Ledger to confirm the issue

spalmer25:
fixes #43 
The `$URL/events?currentscreenonly=true` service does not seem to correspond to what is actually on the screen.
Fortunately, the `$URL/events` service allows the recovery of everything that has been displayed, including what has been lost by `$URL/events?currentscreenonly=true`.

### DIFF
--- a/tests/integration/check_section_text.py
+++ b/tests/integration/check_section_text.py
@@ -47,12 +47,11 @@ class Screen:
     content = content.lstrip('\n')
     return content
 
-def with_retry(url, f, timeout=TIMEOUT):
+def with_retry(f, timeout=TIMEOUT):
     attempts = timeout / 0.5
     while True:
         try:
-            on_screen = get_screen(url)
-            return f(on_screen)
+            return f()
         except AssertionError as e:
             print('- retrying:', e)
             if attempts <= 0:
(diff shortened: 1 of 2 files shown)


tests/integration/check_section_text.py
```diff
@@ -71,6 +71,22 @@ def get_screen(url):
...
+    """Retrieve the screen until title"""
+    r = requests.get(f'{url}/events')
+    r.raise_for_status()
+    lines = [e["text"] for e in r.json()["events"]]
+
+    assert len(lines) > 0, "Unexpected empty screen. Speculos killed?"
+
+    assert title in lines, f"Title \"{title}\" no found"
+
+    title_index = len(lines) - lines[::-1].index(title) - 1
```
emturner, line 84: it would be possibly nice to have this in it's own function, `find_last`?

Possibly could be worth rewriting it to be slightly more efficient too - e.g.
```
def find_last(lst, elm):
  gen = (len(lst) - 1 - i for i, v in enumerate(reversed(lst)) if v == elm)
  return next(gen, None)
```
This avoids copying the entire list :) (from [here](https://stackoverflow.com/questions/6890170/how-to-find-the-last-occurrence-of-an-item-in-a-python-list))

emturner (commented):
Thanks for the fix! One minor comment

emturner (approved):

PR #55, opened 2023-10-10
Tests: add test for all commands

### LINKED ISSUE #54: Test: test all apdu code
For the moment, only the `SIGN_WITH_HASH`,  `SIGN` and `PROMPT_PUBLIC_KEY` commands are tested.
We need to test all the other commands for each device.

spalmer25:
fixes #54 
- [x] Tests for `VERSION`
- [x] Tests for `GIT`
- [x] Tests for `GET_PUBLIC_KEY`
- [x] Tests for unimplemented commands

### DIFF
--- a/docker/Dockerfile.integration-tests
+++ b/docker/Dockerfile.integration-tests
@@ -1,7 +1,7 @@
 FROM ghcr.io/ledgerhq/speculos:sha-bced779
 
 RUN apt update &&                                               \
-    apt install -y curl jq build-essential libsodium-dev git &&	\
+    apt install -y curl jq build-essential libsodium-dev git xxd &&	\
     pip install base58 pytezos pytesseract &&                   \
     git clone --depth 1 --branch v1.11.4 https://github.com/ledgerHQ/ragger && \
     cd ragger &&                                                \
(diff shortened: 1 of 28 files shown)


docker/Dockerfile.integration-tests
```diff
@@ -1,7 +1,7 @@
 FROM ghcr.io/ledgerhq/speculos:sha-bced779
 
 RUN apt update &&                                               \
-    apt install -y curl jq build-essential libsodium-dev git &&	\
+    apt install -y curl jq build-essential libsodium-dev git vim-common &&	\
```
emturner, line 4: why is `vim-common` needed here?

tests/integration/nanos/test_get_pk.sh
```diff
@@ -0,0 +1,9 @@
+start_speculos "$seed"
```
emturner, line 1: could you add a comment explaining which address/its meant to be? (e.g. tz1 vs others). Potentially it would be great to test tz2/tz3 here too?

docker/Dockerfile.integration-tests
spalmer25: It is no longer needed, this must be `xxd` (required for https://github.com/trilitech/ledger-app-tezos-wallet/pull/55/files#diff-f143466a6e0e58c9905d069603b8dc745694c2690409c0b67232f5d1e5fe65ffR51) and not vim-common.
I'm changing that!

emturner (approved):

Issue #58, opened 2023-10-12
EXC_UNEXPECTED_STATE after rejection


spalmer25:
After reject a sign request or the prompt public key command, call any command will raise an `EXC_UNEXPECTED_STATE` exception.
Example:
- Send a `INS_PROMPT_PUBLIC_KEY` (ex: `8003000011048000002c800006c18000000080000000`)
- Reject it
- Retreive a `EXC_REJECT` (=6985)
- Send a `INS_GET_PUBLIC_KEY` (ex: `8002000011048000002c800006c18000000080000000`)
- **Currently**: Retreive a `EXC_UNEXPECTED_STATE` (=9001)
- **Expected**: Retreive the public_key

It seems that after rejection, the `global.step` does not return to `ST_IDLE` but remains at `ST_ERROR`.

PR #59, opened 2023-10-12
Tests: testing rejections/errors workflow

### LINKED ISSUE #57: Tests: testing rejections/errors workflow
The majority of tests are about success scenarios and very few are about the rejections and errors that are triggered.

### LINKED ISSUE #66: Stax: ensure ERR_TOO_DEEP/ERR_TOO_LARGE handled correctly

spalmer25:
### Mandatory tests
fixes #57
fixes #66 
blocked by #58 
 - [x] rejection
 - [x] class exception
 - [x] instruction exception
 - [x] p1 (index) exception
 - [x] p2 (derivation type) exception
 - [x] data length exception
 - [x] wrong data
   - [x] wrong derivation path
   - [x] wrong operation (parse error)
   
### Supplementary tests
- Mixing commands: to detect whether a command is being sent while another is in progress.


tests/integration/nanos/test_reject_transaction.sh
```diff
@@ -0,0 +1,30 @@
+# full input: 0300000000000000000000000000000000000000000000000000000000000000006c016e8874874d31c3fbd636e924d5a036a43ec8faa7d0860308362d80d30e01000000000000000000000000000000000000000000ff02000000020316
+# full output: CAR
+# signer: tz1dyX3B1CFYa2DfdFLyPtiJCfQRUgPVME6E
+# path: m/44'/1729'/0'/0'
+start_speculos "$seed"
+expect_full_text 'ready for' 'safe signing'
+send_async_apdus \
+	800f000011048000002c800006c18000000080000000 "expect_apdu_return 9000
+"\
```
emturner, line 9: could I ask why this slightly odd formatting (multiple places)?

would `"expect_apdu_return 9000" \` work?

app/src/app_main.c
```diff
@@ -127,6 +127,7 @@ app_main(void)
     global.step = ST_ERROR;
 
     for (;;) {
+        TZ_PREAMBLE(("void"));
```
emturner, line 130: feel free to update copyright in this file :)

app/src/keys.c
```diff
@@ -59,7 +59,10 @@ read_bip32_path(bip32_path_t *out, const uint8_t *in, size_t in_size)
     TZ_ASSERT(EXC_WRONG_LENGTH_FOR_INS,
               buffer_read_u8(&cdata, &out->length)
                   && buffer_read_bip32_path(
-                      &cdata, (uint32_t *)&out->components, out->length));
+                      &cdata, (uint32_t *)&out->components, out->length)
+                  // Assert all as been read correctly
```
emturner, line 63:
```
// Assert entire bip32_path consumed
```

tests/integration/stax/utils.py
```diff
@@ -134,6 +136,13 @@ def review_skip_to_signing(self, with_loading=False):
         if with_loading: self.assert_screen("loading_operation")
         self.review.client.resume_ticker()
 
+    @contextmanager
```
emturner, line 139: this is really nice! I guess we could re-use this in the other `review_*` functions as well (future PR)

tests/integration/stax/snapshots/invalid_tag.png
emturner, line 1: this PR will also fix #66 I guess

emturner (commented):
Overall LGTM, great work! 

A couple of ideas:

it would be great if we could define the error codes in `test_runtime.sh` maybe? rather than having special hard-coded hex everywhere, would be awesome to have `expect_apdu_return $ERR_*` potentially? might help to make the tests more readable/understandable.

As a follow up, I think we've definitely got to the point where it would be good to abstract these tests for the nanos devices slightly - e.g. if we introduce `expect_home` and change `expect_section_content` to use the `$TARGET` env var internally, we could de-duplicate *a lot* of these tests, to avoid needing to always add three copies

spalmer25:
> it would be great if we could define the error codes in `test_runtime.sh` maybe? rather than having special hard-coded hex everywhere, would be awesome to have `expect_apdu_return $ERR_*` potentially? might help to make the tests more readable/understandable.

Yes, that's a great idea!

> As a follow up, I think we've definitely got to the point where it would be good to abstract these tests for the nano devices slightly - e.g. if we introduce `expect_home` and change `expect_section_content` to use the `$TARGET` env var internally, we could de-duplicate _a lot_ of these tests, to avoid needing to always add three copies

It's an excellent idea, but I'll deal with it at another time: #67

tests/integration/nanos/test_reject_transaction.sh
spalmer25: Yes, the 2 work perfectly.
Some of the tests were written using the test generator, which should have this offset.
I'll deal with this in another MR: #68

emturner (approved):

PR #64, opened 2023-10-16
Fix EXC_UNEXPECTED_STATE after rejection

### LINKED ISSUE #58: EXC_UNEXPECTED_STATE after rejection

spalmer25:
fixes #58

### DIFF
--- a/app/src/apdu_pubkey.c
+++ b/app/src/apdu_pubkey.c
@@ -112,6 +112,7 @@ static void
 cancel_callback(void)
 {
     stream_cb(TZ_UI_STREAM_CB_REJECT);
+    global.step = ST_IDLE;
     nbgl_useCaseStatus("Address rejected", false, ui_home_init);
 }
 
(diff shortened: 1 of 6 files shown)


tests/integration/stax/test_regression_continue_after_reject.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
```
emturner, line 1: what does the test do that isn't covered by `test_reject_review_operation`? Could adjust that test instead/rename?

spalmer25: Indeed the test `test_reject_review_operation` was already checking this for operation signing.
I remove the test `test_regression_continue_after_reject` from Stax tests and complete the test `test_provide_pk_reject` for the public key prompt.

emturner (approved):

PR #69, opened 2023-10-18
Merge nano tests

spalmer25:


.github/workflows/build.yml
```diff
@@ -114,74 +116,35 @@ jobs:
...
       - name: Create dirs
         run: |
-          mkdir -p tests/samples/micheline/nanos
-          mkdir -p tests/samples/operations/nanos
-
-          mkdir -p tests/samples/micheline/nanosp
-          mkdir -p tests/samples/operations/nanosp
-
-          mkdir -p tests/samples/micheline/nanox
-          mkdir -p tests/samples/operations/nanox
```
emturner, line 124: merging the generated tests makes sense I think 👍🏻 . In which case it might be worth generating some more? but then splitting them between the three devices potentially

tests/generate/Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: nanos nanosp nanox
+.PHONY: nano stax
```
emturner, line 1: would need to modify the below `nanos` etc targets?

tests/integration/nanos/regression_batched_ops.sh
```diff
@@ -1,44 +0,0 @@
-# Regression: second transaction Entrypoint/Parameter not displayed
```
emturner, line 1: yeah, these sorts of tests would be nice to keep around in the specific directories

.github/workflows/build.yml
spalmer25: I think it's better to run the same tests on each device to make sure that a test fails on a specific device and if we wanted to run more tests, we'd have to do it at a different time from the merge requests CI.
What do you think about it?

spalmer25: If you're up for it, I'd like to associate a specific outcome with this change

emturner: I think that makes sense - let's defer until after we submit the app

emturner (approved):

PR #76, opened 2023-10-24
Set up baking app boilerplate

### LINKED ISSUE #75: Baking: 'fork' app-boilerplate codebase into `app_baking`
We will develop the baking app on top of the `app_boilerplate` template. (can be found [here](https://github.com/LedgerHQ/app-boilerplate)). 

Will need to integrate the 'build' with CI here.

# Requirements
- #74

spalmer25:
fixes #75


app_baking/Makefile
```diff
@@ -0,0 +1,108 @@
+# ****************************************************************************
```
emturner, line 1: Not copying license is good - checked and agree that we use exactly the same license here.

app_baking/Makefile
```diff
@@ -0,0 +1,108 @@
...
+#    Ledger App Boilerplate
```
emturner, line 2: checked, agree that the code has been verbatim copied across.

app_baking/Makefile
```diff
@@ -0,0 +1,108 @@
...
+#    (c) 2023 Ledger SAS.
```
emturner, line 3: should we keep the `tests` directory actually? Eventually, we may well want to transition the `wallet` tests to use ragger/python for nano devices too.

If we use them within the baking app project, it would be a really good opportunity to learn how they work properly? wdyt?

app_baking/Makefile
```diff
@@ -0,0 +1,108 @@
...
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
```
emturner, line 5: Do we want to keep the fuzzing workflow around too? Sounds like it could be potentially really useful, for e.g. stress testing the `wallet`/`baking` parser in future?

emturner (commented):
overall lgtm, main comment around tests

app_baking/Makefile
spalmer25: Yes, it seems right to keep them.
I removed them because we already had the tests/ folder.
But it does make sense to keep them while we merge them with the tests/ folder.

emturner (approved):
LGTM

tests/fuzzing to be enabled in CI by a follow up

emturner (approved):
LGTM

PR #94, opened 2023-10-25
stax: fix home text

### LINKED ISSUE #84: Stax: improve home screen text
“Ready for signing” should be “This app enables signing transactions on the Tezos network”

emturner:
Update Stax home screen text according to Ledger's request.

- closes #84


app/src/ui_home_nbgl.c
```diff
@@ -107,21 +107,24 @@ ui_toggle_clear_blind(void)
...
         const char *button_text;
         const char *tagline;
 
         switch (global.home_screen) {
         case SCREEN_CLEAR_SIGN:
-            tagline     = "Ready for clear signing";
+            tagline     = CLEAR_SIGN_HOME_TEXT;
             button_text = "blind sign";
             break;
         case SCREEN_BLIND_SIGN:
```
spalmer25, line 130: Don't we want a similar text for SCREEN_BLIND_SIGN?

spalmer25:
Just one more comment: our app doesn't just sign transactions, it signs several types of operation.
Wouldn't it be more appropriate to have the text "This app enables signing operations on the Tezos Network"?
Or does the term "transaction" refer to all these types of operation?

app/src/ui_home_nbgl.c
emturner: Makes sense, may as well.

emturner:
> Just one more comment: our app doesn't just sign transactions, it signs several types of operation. Wouldn't it be more appropriate to have the text "This app enables signing operations on the Tezos Network"? Or does the term "transaction" refer to all these types of operation?

I'm not sure - Ledger asked for this text (I did raise, but haven't heard back yet). I suspect that it might be more geared towards new users? ie new users are probably only really doing transfers. More advanced users would know the difference, when they start reaching for other operations

app/src/ui_home_nbgl.c
emturner: Done

spalmer25 (approved):

spalmer25:
LGTM !
Just in case you haven't noticed, the base branch is still on `emturner@tooling` and should be moved back to `main`.

PR #99, opened 2023-10-26
CI: fix docker rebuild on every merge to main

emturner:
See https://github.com/trilitech/ledger-app-tezos-wallet/actions/runs/6656361822/job/18088730148 for a successful run.

Previously docker images were getting built every time a merge to master occurs


spalmer25 (approved):
Thank you very much!

Issue #80, opened 2023-10-25
Nano: remove gaps between title & icon, title below icon


emturner:
# Goal
When displaying title + icon, the icon should appear immediately above the title, and there should be a blank line either side.
The right/left arrows should appear in the middle of the screen, when on the home screen.

## Current state

![screenshot-2023-10-25-12:01:49](https://github.com/trilitech/ledger-app-tezos-wallet/assets/36469783/4f937b90-5cd0-41ac-ac4b-00fc66951595)

## Target state

![screenshot-2023-10-25-12:02:18](https://github.com/trilitech/ledger-app-tezos-wallet/assets/36469783/2c64b064-aa30-43f2-81fc-2b227d04ff62)

spalmer25:
Does 
> When displaying title + icon, the icon should appear immediately above the title

also apply to Nanos device?

PR #101, opened 2023-10-27
Update approve text

### LINKED ISSUE #86: Nano: update text on signing `approve` screen
# Current
screenshot-2023-10-25-14:26:29

# Desired

screenshot-2023-10-25-14:24:49

spalmer25:
fixes #86
requires #109 
 - On Nano(sp/x): `Accept?\nPress both buttons to accept.` -> `Accept\nand send`
 - On Nanos: `Accept?` -> `Accept and send`


emturner (commented):
overall LGTM, have that same issue with the second line of title being blank (as with PK MR)

screenshot-2023-10-31-14:53:02

ajinkyaraj-23:
I manually tested for nanox and nanosp using speculos. "Accept \n and send" working fine on both of them.

app/src/ui_stream.c
```diff
@@ -361,6 +365,22 @@ redisplay(void)
...
+redisplay(void)
+{
+    FUNC_ENTER(("void"));
+
+    tz_ui_stream_t *s      = &global.stream;
+    size_t          bucket = s->current % TZ_UI_STREAM_HISTORY_SCREENS;
+
+    // clang-format off
+    switch (s->screens[bucket].layout_type) {
+    case TZ_UI_LAYOUT_BNP: redisplay_bnp(); break;
```
emturner, line 462: do we need a `default: TZ_FAIL(EXC_UNKNOWN)`?

(might need a switch to the `exception.h` preamble/post amble here

app/src/ui_stream.c
```diff
@@ -374,6 +365,107 @@ redisplay(void)
...
+        if (s->current == s->total - TZ_UI_STREAM_HISTORY_SCREENS + 1)
+            init[1].text = (const char *)&C_icon_go_forbid;
+        else
+            init[1].text = (const char *)&C_icon_go_left;
+    }
+    /* If we aren't full or aren't on the last page, we can go right */
+    if (!s->full || s->current < s->total)
+        init[2].text = (const char *)&C_icon_go_right;
+
+    DISPLAY(init, cb);
```
emturner, line 448: quite a lot of this seems identical to `redisplay_bnp`? Is it possible to keep the repeated parts identical between the two (not sure if it'll be de-duplicated in future?)

But the logic about whether or not to display back/forward arounds seems like it could be abstracted out?

emturner (commented):
overall LGTM, couple questions

app/src/ui_stream.c
spalmer25: Yes, there is quite a lot of duplication.
As this isn't really the aim of the PR, I thought it would be better to deal with it in issue #80, since it deals entirely with the way screens are displayed.

app/src/ui_stream.c
spalmer25: Yes, you're right, it could be helpful!

emturner (approved):

PR #105, opened 2023-10-30
Stax: review operation splash screen must wait for tap to advance

### LINKED ISSUE #89: Stax: 'review request to sign' splash screen must only advance with finger press
# Current behaviour

The 'review request' splash screen is briefly displayed, before immediately advancing to the first page of the operation, on receipt of the follow up APDU.

To reproduce:

- start STAX app in speculos
- send the apdu `800f000011048000002c800006c18000000080000000`
- observe the `Review request to` splash screen (if using the webview, you may need to refresh)
- send the apdu `800f81005d0300000000000000000000000000000000000000000000000000000000000000006b00ffdd6102321bc251e4a5190ad5b12b251069d9b4904e02030400747884d9abdf16b3ab745158925f567e222f71225501826fa83347f6cbe9c393` 
- observe the screen changes to `Operation (0)` (if using the webview, you may need to refresh)

# Goal

The screen should only advance on a user press, even if an apdu packet is received in the meantime.

emturner:
- Fixes #89 
- realised this same issue kind of also applies to the parsing error screens, which were shown immediately, rather than waiting for the user to reach them


tests/integration/stax/utils.py
```diff
@@ -136,6 +136,11 @@ def review_skip_to_signing(self, with_loading=False):
         if with_loading: self.assert_screen("loading_operation")
         self.review.client.resume_ticker()
 
+    def review_tap(self, screenShouldChange = True):
+        self.review.tap()
+        if screenShouldChange:
+            self.__backend.wait_for_screen_change()
+
```
spalmer25, line 143: What is the purpose of this function?
I haven't seen it used anywhere?

tests/integration/stax/test_parsing_errors.py
```diff
@@ -15,6 +17,21 @@
 
 from utils import *
 
+def exit_err_screen(app, code):
+    app.review.tap()
```
spalmer25, line 21: This tap is not usefull.
`nbgl_useCaseStatus` draw a transient status page, without control

tests/integration/stax/utils.py
emturner: Oh 🤦🏻‍♀️ - thanks, I didn't need this in the end, will remove

tests/integration/stax/test_parsing_errors.py
emturner: 👍🏻 moving `app.review.tap()` to `review_reject_screen` - used to exit early

spalmer25 (approved):

PR #113, opened 2023-11-03
wallet: Add public key to globals

### LINKED ISSUE #108: prompt address: only derive PK once
# Goal
When handling `INS_PROMPT_PUBLIC_KEY`, we should only derive the public key once.

# Context

              One thing spotted: we derive pk twice in the prompt flow!

_Originally posted by @emturner in https://github.com/trilitech/ledger-app-tezos-wallet/issues/106#issuecomment-1789387894_

ajinkyaraj-23:
fixes #108. 
Add pk to globals to save repeated computation of pk. added pub key to global.key in union with apdu as apdu.hash/sign is not used concurrently with pk in any operation.

### DIFF
--- a/app/src/apdu_pubkey.c
+++ b/app/src/apdu_pubkey.c
@@ -35,25 +35,21 @@
 
 static void provide_pubkey(void);
 static void prompt_address(void);
-static void format_pkh(char *, size_t);
+static void format_pkh(cx_ecfp_public_key_t *, char *, size_t);
 static void stream_cb(tz_ui_cb_type_t);
 
 static void
 provide_pubkey(void)
 {
-    buffer_t             bufs[2] = {0};
-    uint8_t              byte;
-    cx_ecfp_public_key_t pubkey;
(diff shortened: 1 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -10,3 +10,9 @@ repos:
       - id: end-of-file-fixer
       - id: check-yaml
       - id: check-added-large-files
+  - repo:  https://github.com/pre-commit/mirrors-clang-format
```
emturner, line 13: please remove - this breaks everything on my machine 😅

Makefile
```diff
@@ -135,10 +135,6 @@ load_%: app_%.tgz
 #
 # Dash vs under aliases:
 
-format:
```
emturner, line 138: please restore 😅 
(also - formatting probably shouldn't be part of this commit)

emturner (approved):
LGTM - confirmed that this doesn't increase size of `globals`

Issue #91, opened 2023-10-25
Stax: display multiple values per screen where possible.


emturner:
We should display multiple values on the screen where possible (e.g. Fee, Amount and Destination could all easily fit on the same screen).

Currently, only a maximum of one value is displayed on each screen.

- requires #23

Issue #23, opened 2023-09-26
Stax: handle screen text properly


emturner:
# Goal

Implement proper line/pixel based screen text for stax, in the same style as the nano devices. 

On large values, as much of the screen as possible should be used.

Additionally, where very large values are present, we should display them in a `details` modal

# Context

Currently, the STAX app draws content as a hardcoded '60' characters, which seems to usually take up roughly half a screen. This is done to simplify the initial clearsigning logic.

See: https://github.com/trilitech/ledger-app-tezos-wallet/pull/18#discussion_r1336084377 for context.

PR #115, opened 2023-11-07
ui_strings: a ring buffer for screen text

emturner:
progress towards:
- #91 
- #23 

# Context
We are trying to update STAX to allow multiple values per screen. Doing so in the same method  as `ui_stream.c` would be very wasteful though - as stax values could be _up to_ well over 120+ bytes per value on a single screen, but most of the time we don't need anything close to that (e.g. `Fee` field etc).

Therefore, we compress all strings in memory, into a 'sudo'-ring-buffer, to allow more efficient use memory, for screens (on all devices) that consume fewer bytes than are pre-allocated arrays in the `global.screens[bucket]`.

This actually results in a saving of well over 30bytes on the nanos, and with the current parameters, over 250-300bytes on the nanosp/x. It will theoretically save potential 1-2 kilobytes on the Stax, if we implement history there (we currently have 8 history screens there currently, so it already will save a lot of memory, but we don't currently make use of them).

### DIFF
--- a/app/src/exception.h
+++ b/app/src/exception.h
@@ -20,8 +20,8 @@
 
 #pragma once
 
-#include <os.h>
 #include <io.h>
+#include <os.h>
 
 #define SW_OK 0x9000
 
--- a/app/src/ui_strings.c
+++ b/app/src/ui_strings.c
@@ -1,0 +1,330 @@
+/* Tezos Ledger application - Generic stream display
+
+   Copyright 2023 TriliTech <contact@trili.tech>
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License. */
+
+#include <io.h>
+#include <stdbool.h>
+
+#include "globals.h"
+#include "ui_strings.h"
+
+/* Allow future unit tests to override */
+#ifndef UI_STRINGS
+static tz_ui_strings_t ui_strings;
+#define UI_STRINGS &ui_strings
+#endif
+
+#define BUFF_START ((char *)(s->buffer))
+#define BUFF_END   ((char *)(s->buffer) + BUFF_LEN)
+
+/* Prototypes */
+void   ui_strings_init(void);
+void   ui_strings_push(const char *str, size_t len, char **out);
+void   ui_strings_drop(char **str);
+size_t ui_strings_fit_up_to(size_t len, char **write_start);
+void   ui_strings_can_fit(size_t len, bool *can_fit);
+bool   ui_strings_is_empty();
+
+#ifdef TEZOS_DEBUG
+void ui_strings_print();
+#define PRINT_STRINGS ui_strings_print()
+#else
+#define PRINT_STRINGS
+#endif
+
+/* Definitions */
+void
+ui_strings_init(void)
+{
+    tz_ui_strings_t *s = UI_STRINGS;
+
+    memset(s->buffer, '\0', BUFF_LEN);
+    s->start        = BUFF_START;
+    s->end          = BUFF_START;
+    s->internal_end = BUFF_START;
+    s->count        = 0;
+}
+
+/* @param len: we want to a string s of strlen(s) == len
+   @param write_start: this is where we'll start writing this string in the
+   buffer
+   @return out_len: this is the length of the string we can fit. out_len <=
+   len
+*/
+size_t
+ui_strings_fit_up_to(size_t len, char **write_start)
+{
+    tz_ui_strings_t *s       = UI_STRINGS;
+    size_t           out_len = 0;
+    TZ_PREAMBLE(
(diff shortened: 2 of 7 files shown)


app/src/exception.h
```diff
@@ -20,6 +20,7 @@
 
 #pragma once
 
+#include <io.h>
```
ajinkyaraj-23, line 23: Isn't it duplicate?

app/src/ui_strings.c
```diff
@@ -0,0 +1,320 @@
...
+#include "globals.h"
+#include "ui_strings.h"
+
+#define BUFF_START ((char *)(s->buffer))
+#define BUFF_END   ((char *)(s->buffer) + BUFF_LEN)
+
+void
+ui_strings_init(void)
+{
+    tz_ui_strings_t *s = &global.stream.strings;
```
ajinkyaraj-23, line 29: provide the global.stream.strings pointer as argument to init function, which removes dependency of u_strings library on globals. It makes it easier to unit test the library standalone. I checked that globals is not used for any thing else in ui_strings.c/h.
Supply the tz_ui_strings_t *s as static variable in header file, which is assigned to the argument in init function.

app/src/ui_strings.c
```diff
@@ -0,0 +1,320 @@
...
+        TZ_SUCCEED();
+    }
+
+    /* start < end */
+    size_t bytes_at_start = MIN((size_t)(s->start - BUFF_START), len + 1);
+    size_t bytes_at_end   = MIN((size_t)(BUFF_END - s->end), len + 1);
+
+    if (bytes_at_end > len || bytes_at_end >= bytes_at_start) {
+        *write_start = s->end;
+        *write_end   = s->end + bytes_at_end - 1;
```
ajinkyaraj-23, line 96: we can remove write_end and just return bytes_at_end/bytes_at_start as length. 
If you return it like out_len = bytes_at_start (or _end depending on the if else), then in ui_strings_can_fit you can just check out_len > Len. you don't need do -1 here and then add >= at line 117. 
In the push and drop functions also we have effectively used ws and len. This will kind of abstract out the null character calculation while reading the code.

app/src/exception.h
emturner: I found that I had to include this, otherwise just including `"exception.h"` alone gives you compilation errors

app/src/ui_strings.c
emturner: I'd rather keep it as is for now, since we've used almost exactly this pattern everywhere else. That being said, I'm happy to replace it with something like the following in the declaration section of `ui_strings.c`:

```
#ifndef UI_STRINGS
#define UI_STRINGS &global.stream.strings
#endif
...
tz_ui_strings_t *s = UI_STRINGS;
```

Which would allow a future unit test file to override this?
What do you think?

app/src/ui_strings.c
emturner: Quite possibly - will take a look

app/src/exception.h
ajinkyaraj-23: Can you remove line 25. we are including io.h twice.

emturner: Ah sorry, think that's a rebase artifact. Will remove

emturner: done

app/src/ui_strings.c
emturner: done. Added in first commit to point to static, updated in second to remove static and point to global

app/src/ui_strings.c
emturner: done

app/src/ui_strings.c
```diff
@@ -0,0 +1,330 @@
...
+        out_len      = len;
+
+        TZ_SUCCEED();
+    }
+
+    /* Buffer not empty */
+
+    if (s->start > s->end) {
+        *write_start = s->end;
+        out_len      = (s->start - 1 - s->end);
```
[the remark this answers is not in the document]
emturner: ah thanks - good point

app/src/ui_strings.c
```diff
@@ -0,0 +1,330 @@
...
+ui_strings_can_fit(size_t len, bool *can_fit)
+{
+    char  *ws = NULL;
+    size_t out_len;
+
+    TZ_PREAMBLE(("len=%d", len));
+
+    TZ_CHECK(out_len = ui_strings_fit_up_to(len, &ws));
+
+    *can_fit = (out_len >= len);
```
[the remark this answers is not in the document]
emturner: done

app/src/ui_strings.c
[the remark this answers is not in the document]
emturner: done (equivalent)

app/src/ui_strings.c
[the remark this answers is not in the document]
emturner: done

app/src/ui_strings.c
```diff
@@ -0,0 +1,330 @@
...
+    /* Internal checks */
+    TZ_ASSERT(EXC_MEMORY_ERROR, (s->end <= s->internal_end));
+
+    char  *ws = NULL;
+    size_t out_len;
+    TZ_CHECK(out_len = ui_strings_fit_up_to(len, &ws));
+    PRINTF("[DEBUG] Found space from %p to %p (for %d chars)\n", ws,
+           ws + out_len, out_len);
+
+    TZ_ASSERT(EXC_MEMORY_ERROR, (out_len >= len));
```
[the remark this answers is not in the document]
emturner: done

ajinkyaraj-23 (approved):

PR #116, opened 2023-11-08
wallet: local clang-format added to pre-commit.

ajinkyaraj-23:


.github/workflows/pre-commit.yml
```diff
@@ -10,4 +10,6 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '3.10'
+      - name: Install clang-format
+        run:  sudo apt-get install clang-format-15
```
emturner, line 14: is there a way to avoid it running the `clang-format` check in CI?

formatting is already checked by one of the other workflows (possibly guidelines_enforcer?) Would be good to avoid doing duplicated work in CI when we can avoid it

ajinkyaraj-23: We are not running clang-format-15. We are just installing it. 
clang-format is run as part of pre-commit on local machine during development. It can not be removed from there. The same pre-commit file is being used in this workflow.  So it will be there. 
Also in guideline enforcer we are using clang-static analyser which is different from clang-format. 
This check here is useful when someone makes a commit directly on GitHub or forces a commit while ignoring to run pre-commit.

ajinkyaraj-23: Skipped clang-format while running pre-commit ci workflow.

.github/workflows/pre-commit.yml
```diff
@@ -10,4 +10,9 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '3.10'
-      - uses: pre-commit/action@v3.0.0
+      - run: python -m pip install pre-commit
+        shell: bash
+      - run: python -m pip freeze --local
+        shell: bash
+      - run: SKIP=clang-format pre-commit run --all-files
+        shell: bash
```
emturner, line 18: you probably don't have to specify `shell: bash` here? I think it's the default

emturner (approved):
otherwise LGTM

PR #117, opened 2023-11-09
Tests: use snapshots for basic tests for nano devices

### LINKED ISSUE #111: Nano: use snapshots for basic tests
At the moment, all the basic tests for `nano` devices use text comparison to ensure that the screen is correct.
But this type of comparison cannot detect display changes.

By using snapshots, the tests can guarantee that the screen content and display have not changed.

spalmer25:
fixes #111


tests/integration/nano/utils/app.py
```diff
@@ -0,0 +1,152 @@
+# Copyright 2023 Functori <contact@functori.com>
```
emturner, line 1: V minor - should probably have the trili one here too? (Think some of this stuff has been lifted from the stax utils.py?)

As a follow up - this awesome! Potentially would be able to reimplement the `stax` one to inherit from here.

tests/integration/nano/test_basic.py
```diff
@@ -0,0 +1,48 @@
...
+        app.assert_screen(Screen.Home)
+        app.backend.right_click()
+        app.assert_screen(Screen.Version)
+        app.backend.right_click()
+        app.assert_screen(Screen.Settings)
+        app.backend.right_click()
+        app.assert_screen(Screen.Quit)
+        app.backend.right_click()
+        app.assert_screen(Screen.Quit)
+        app._quit()
```
emturner, line 48: why `app._quit` and not `app.quit`?

tests/integration/.gitignore
```diff
@@ -0,0 +1 @@
+/nano/snapshots-tmp
```
emturner, line 1: where does this come from?

tests/integration/nano/test_basic.py
emturner: Oh I see - `quit` starts from home 👍🏻

tests/integration/nano/utils/app.py
spalmer25: Yes, I'm definitely going to add the trili coryright here too!

tests/integration/nano/test_version.py
```diff
@@ -0,0 +1,28 @@
...
+# limitations under the License.
+
+from utils.app import *
+
+if __name__ == "__main__":
+    with nano_app() as app:
+
+        app.assert_screen(Screen.Home)
+
+        data = app.backend.version()
```
emturner, line 23: as an aside - it's currently slightly painful to get `VERSION_BYTES` and `COMMIT` in scope correctly. Would it be possible to adjust so that it's an optional param - and throw if it's requested but unavailable?

That way most of the tests can be manually run/re-snapshotted without needing to load those in each time

emturner: (like, not super painful - currently `export VERSION_BYTES=$VERSION_BYTES` etc fixes it, but it's a bit annoying

tests/integration/nano/test_git.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
```
emturner, line 1: I see git/version tests here - but no snapshots in the same commit?

tests/integration/.gitignore
spalmer25: To enable debugging, `backend.compare_screen_with_snapshot` supports a folder in which temporary snapshots will be stored (so that we can compare the actual snapshot with the expected snapshot).
I use the same folder as the `Navigator` of `ragger` so that all snapshots are centralised.

tests/integration/nano/test_git.py
spalmer25: The git and version tests test app requests which display nothing on the Ledger.
Other tests are in this case, such as tests for wrong packets: https://github.com/trilitech/ledger-app-tezos-wallet/pull/117/commits/853290613e10dec9578165df8d2ef7a079bdd696

tests/integration/nano/test_provide_tz3_pk.py
```diff
@@ -0,0 +1,32 @@
...
+    test_name = Path(__file__).stem
+    with nano_app() as app:
+
+        app.assert_screen(Screen.Home)
+
+        account = Account("m/44'/1729'/0'/0'", SIGNATURE_TYPE.SECP256R1)
+        data = app.provide_public_key(account, test_name)
+
+        app.check_public_key(
+            public_key="410497f4d381101d2908a13669313faec5dbf6693985584f96268ea2c25178199ddd1aad041e7564795eb4b9a4f379e8cdc0c8391f7b2880613771fff76e6a6b05cf",
```
emturner, line 29: would it be possible to do a `from base58 check` instead (converting to hex/bytes internally)? That way it would be human readable :)

emturner: would also open the doors to abstracting the `test_(provide|get)_pk` into one or two functions with a bunch of different example tests

emturner: and make it clear that the pk expected is definitely linked to the address shown

tests/integration/nano/test_version.py
spalmer25: Yes, it's absolutely right!
What about retrieving it directly from the script using `GitPython`?

tests/integration/nano/test_provide_tz3_pk.py
spalmer25: Yes, that could be awesome, I'll have a look!

tests/integration/nano/test_wrong_instructions.sh
```diff
@@ -1,8 +0,0 @@
-start_speculos "$seed"
```
emturner, line 1: I don't see a `.py` test for this one?

tests/integration/nano/utils/app.py
```diff
@@ -335,11 +335,15 @@ def parsing_error_signing(self,
...
+                              text: str,
                               path: Union[str, Path]) -> None:
+        if self.backend.firmware.device != "nanox":
+            text = "Parsing error"
+
         self._failing_signing(account,
                               message,
                               with_hash,
-                              "Parsing error",
+                              text,
```
emturner, line 346: what does this change do? (I think I'm forgetting something 😓 )

emturner (commented):
overall I think this makes sense? It does get rid of the `check_section_text.py` stuff from the  basic tests IIUC - is that likely to be an issue at all?

tests/integration/.gitignore
emturner: Would be awesome to add a section to the `README` perhaps, to talk through how to run tests/reset the snaps?

tests/integration/nano/test_git.py
emturner: > The git and version tests test app requests which display nothing on the Ledger.

The home screen should be visible, no?

spalmer25: The home screen snapshot is common to all the tests and has already been added in a previous commit.

tests/integration/nano/utils/app.py
spalmer25: In fact, the speculo for Nanox does not work correctly (see https://github.com/trilitech/ledger-app-tezos-wallet/issues/43)
We managed to fix this problem in check_section_text.py, but here the check is made in the Navigator.

So to detect the parsing error screen for Nanox I changed the text to be detected.
I will add a comment for that!

tests/integration/nano/test_wrong_instructions.sh
spalmer25: I noticed that this test is included in test_unimplemented_commands.py
So I removed it

tests/integration/nano/utils/app.py
emturner: Oh yeah 🤦🏻‍♀️ - thanks!

emturner (approved):
🔥

PR #119, opened 2023-11-09
stax: display multiple values per page

### LINKED ISSUE #91: Stax: display multiple values per screen where possible.

emturner:
fixes #91 

*NB* there are still a couple artifacts left
- enabling text wrapping will be enabled in a follow up PR
- adding details page will be enabled in a follow up PR (part of #23)


Makefile
```diff
@@ -2,8 +2,8 @@
 # Makefile
 #
 
-all: app_nanos.tgz app_nanosp.tgz app_nanox.tgz
-debug: app_nanos_dbg.tgz app_nanosp_dbg.tgz app_nanox_dbg.tgz
+all: app_nanos.tgz app_nanosp.tgz app_nanox.tgz app_stax_dbg.tgz
```
ajinkyaraj-23, line 5: Why stax_dbg in line all: when others are not dbg.

emturner: this was an artifact that shouldn't have been included. Have removed

app/src/ui_strings.c
```diff
@@ -222,6 +221,43 @@ ui_strings_drop(char **in)
...
+    s->count--;
+
+    s->end = *in;
+
+    *in = NULL;
+
+    if (s->end == BUFF_START && !ui_strings_is_empty()) {
+        s->end = s->internal_end;
+    } else if (s->start <= s->end) {
+        s->internal_end = s->end;
```
ajinkyaraj-23, line 253: Verified that logic for drop last is correct. 
When it's the last string in buffer, start == end == internal_end.

app/src/ui_stream.c
```diff
@@ -534,29 +417,146 @@ tz_ui_stream(void)
...
+    will_fit = se_get_cropped_length(value, will_fit, BAGL_WIDTH,
+                                     BAGL_ENCODING_LATIN1);
+#elif defined(HAVE_BAGL)
+    uint8_t width;
+    will_fit++;
+    do {
+        width = bagl_compute_line_width(BAGL_FONT_OPEN_SANS_REGULAR_11px, 0,
+                                        value, --will_fit,
+                                        BAGL_ENCODING_LATIN1);
+    } while (width >= BAGL_WIDTH);
```
ajinkyaraj-23, line 443: Why we need while loop to calculate line_width. It's straightforward calculation which should be done once.

emturner: Because we're not calculating the line width tself - we're searching for the maximum number of chars that fit in one line.

Which is annoying to do since the text is variable width

spalmer25 (approved):
Overall LGTM

app/src/ui_stream_nbgl.c
```diff
@@ -114,15 +114,17 @@ tz_choice_ui(bool accept)
...
-    FUNC_ENTER(("void"));
+    tz_ui_stream_t *s          = &global.stream;
+    int16_t         prev_total = s->total;
 
-    tz_ui_stream_t *s = &global.stream;
+    TZ_PREAMBLE(("prev_total=%d", prev_total));
 
-    if (!s->full)
-        s->cb(TZ_UI_STREAM_CB_REFILL);
+    if (!s->full && prev_total == s->total) {
```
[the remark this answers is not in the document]
emturner: Oh great catch! I was doing something with `prev_total` at one point, but no more :)

PR #121, opened 2023-11-13
tests/stax: snapshot each test into its own directory

emturner:
Broken out from #119, since it doesn't touch any app code.


emturner:
There are a _lot_ of changes to screens as a result of #119 - this change allows us to actually track a bit easier which flows change as a result.

I'm definitely not against de-duplicating things like `home` in future - but having full separation clears things up initially.

(would probably happen as part of adapting the stax testing framework to reuse the new nano one)

spalmer25:
LGTM

spalmer25 (approved):

PR #124, opened 2023-11-14
Add a review operation screen at start of signing flow

### LINKED ISSUE #123: Nano: add an initial screen on the flow of the operation signature
# Goal
Add a screen similar to: 
Screenshot from 2023-11-14 10-44-33
At the start of the flow of signing operation

spalmer25:
fixes #123


spalmer25:
The new way of testing signing flow uses the `ragger` Navigator which performs a sequence of instructions, taking and comparing screenshots with those expected between each instruction.
In order to take the correct snapshots, the Navigator waits after each instruction for the screen to be changed, i.e. :
- it takes a snapshot of the current screen and compares it with the previous one,
  - if the snapshots are different it moves on to the next instruction,
  - if not, it repeats the comparison after a short time. 
    - If after a certain timeout the snapshots remain identical, it raises an exception.

In the `test_too_deep.py` test with the Nanos device, the to_deep error occurs after displaying 2 screens containing the text `{{{{{{{{{{{{{{{{{{{`.
With luck, the test passed thanks to the left arrow present in the second screen but not in the first (which made the screens different).
But by adding a first screen to the signing flow the 2 screens become identical and so the Navigator will wait until the timeout for a change of screen on the second data screen.

I'm going to modify the `test_too_deep.py` test to make it work but this problem could occur for future tests so this problem needs to be fixed generally.
I will open a new issue for this.

emturner (approved):
lgtm

Re the 'same screens' issue - I wonder if we should enable a `screen counter` in the title - for the number of screens that we've seen with the same title?

That way, all the screens would be guaranteed to be different

emturner (approved):

PR #126, opened 2023-11-14
stax: enable text wrapping

emturner:
part of #23 

We enable wrapping - so that words are not split across multiple lines where possible.


ajinkyaraj-23 (approved):
LGTM

spalmer25 (approved):

PR #130, opened 2023-11-16
docker: more precise pinning of packages for building tests image

emturner:
Failing to build currently due to timeout during dependency resolution.


ajinkyaraj-23 (approved):
It works for me too. Thanks

PR #145, opened 2023-11-27
stax: append value to previous when titles the same

### LINKED ISSUE #23: Stax: handle screen text properly

emturner:
close #23 

This aims to use up more screen <most of the time>. We reduce the size of the 'parser buffer', and construct screens more incrementally.


app/src/ui_stream_nbgl.c
```diff
@@ -351,18 +352,34 @@ tz_ui_stream_pushl(tz_ui_cb_type_t cb_type, const char *title,
...
-        push_str(&value[offset], length,
-                 (char **)&s->screens[bucket].pairs[idx].value);
+        char *out = NULL;
 
-        offset = length;
+        if (append) {
+            bool can_fit = false;
+            ui_strings_can_fit(length, &can_fit);
+            if (!can_fit)
+                drop_last_screen();
```
spalmer25, line 374: What happens if, after `drop_last_screen()`, `length` still `!can_fit`?
Is this possible?

tests/integration/stax/snapshots/test_parsing_errors/tpe_review_too_deep_0.png
spalmer25, line 1: I'm not sure why this screen is being displayed now

emturner: Because originally, the error showed up before the parser had produced 60 chars of output. 

But now, with the parsing buffer reduced back down, the first 20 chars of output are apparently produced before encountering the error.

When an error is encountered, we mark the current screen as complete and push the error to the next screen.

The smaller parsing buffer allows us to fill up more of the screen on average, for larger fields

spalmer25: Awesome

spalmer25 (approved):

PR #146, opened 2023-11-28
Fix operation parser READ_INT32 step

spalmer25:
The `READ_INT32` step of the operation parser does not respect byte-by-byte parsing.
This PR:
 -  fix it
 - modifies the unit-tests to detect non-respect of byte-by-byte parsing
 - allows the app to receive several packets before waiting for user input


app/src/parser/operation_state.h
```diff
@@ -125,6 +125,10 @@ typedef struct {
             tz_operation_field_kind kind : 5;
             uint8_t                 skip : 1, natural : 1;
         } step_read_num;
+        struct {
+            uint32_t value;
+            uint8_t  skip : 1, ofs : 3;
```
emturner, line 130: is this (I'm assuming) applying default values to the skip/ofs fields of the struct?

emturner: Why is `skip = 1` and not `0`?

tests/integration/nano/test_sign_with_small_packet.py
```diff
@@ -0,0 +1,37 @@
...
+from utils.app import *
+
+if __name__ == "__main__":
+    test_name = Path(__file__).stem
+    with nano_app() as app:
+
+        def check_byte_by_byte(message: str, signature: str, path: str) -> None:
+            app.assert_screen(Screen.Home)
+
+            data = send_and_navigate(
```
emturner, line 26: should this produce snapshots at all?

spalmer25: `app.navigate_until_text` will produce snapshots and compare them with those in the `test_sign_transaction` folder.
This folder already exists because it's the one that contains the `test_sign_transaction` test snapshots.
But, on reflection, this isn't a good idea because their `snapshot-tmp` will be stored in the same place.
I'm going to change that

app/src/parser/operation_state.h
spalmer25: `uint8_t skip: 1, ofs: 3;` means that `skip` will be stored in `1` bit and `ofs` in `3` bits (because `ofs` must not exceed `4`, `100` in binary).
https://www.geeksforgeeks.org/bit-fields-c/

emturner: oooh, thanks so much!

emturner (approved):

PR #159, opened 2023-12-07
Add complex trait to operation fields

### LINKED ISSUE #158: Tag too complex fields in the parser
Some operation fields can be considered too complex for a standard user to understand.
(See Ledger's remark https://docs.google.com/document/d/1XJYhHh9lMtgqtsKbCjlcxCrHzurrDR1GXYSMOV5VTh0/edit#heading=h.1rgdf7dp5gu5)
Objectf:
Have a way to detect fields that are too complex from the parser.

spalmer25:
fixes #158 
This PR updates the parser to keep track of the complexity of the current operation field.

### DIFF
--- a/app/src/parser/operation_parser.c
+++ b/app/src/parser/operation_parser.c
@@ -49,175 +49,176 @@ const char *const tz_operation_parser_step_name[] = {"MAGIC",
                                                      "READ_PROTOS"};
 #endif
 
+// clang-format off
 const tz_operation_field_descriptor proposals_fields[] = {
-  // Name,       Kind,                      Req,  Skip,  None
-    {"Source",   TZ_OPERATION_FIELD_PKH,    true, false, false},
-    {"Period",   TZ_OPERATION_FIELD_INT32,  true, false, false},
-    {"Proposal", TZ_OPERATION_FIELD_PROTOS, true, false, false},
-    {NULL,       0,                         0,    0,     0    }
+  // Name,       Kind,                      Req,  Skip,  None,  Complex
+    {"Source",   TZ_OPERATION_FIELD_PKH,    true, false, false, false},
+    {"Period",   TZ_OPERATION_FIELD_INT32,  true, false, false, false},
+    {"Proposal", TZ_OPERATION_FIELD_PROTOS, true, false, false, false},
+    {NULL,       0,                         0,    0,     0,     0    }
 };
 
 const tz_operation_field_descriptor ballot_fields[] = {
-  // Name,       Kind,                      Req,  Skip,  None
-    {"Source",   TZ_OPERATION_FIELD_PKH,    true, false, false},
-    {"Period",   TZ_OPERATION_FIELD_INT32,  true, false, false},
-    {"Proposal", TZ_OPERATION_FIELD_PROTO,  true, false, false},
-    {"Ballot",   TZ_OPERATION_FIELD_BALLOT, true, false, false},
-    {NULL,       0,                         0,    0,     0    }
+  // Name,       Kind,                      Req,  Skip,  None,  Complex
+    {"Source",   TZ_OPERATION_FIELD_PKH,    true, false, false, false},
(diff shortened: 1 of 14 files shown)


app/src/parser/operation_parser.c
```diff
@@ -709,8 +708,10 @@ tz_operation_parser_step(tz_parser_state *state)
             uint8_t present = 1;
             if (!field->required)
                 tz_must(tz_parser_read(state, &present));
-            if (!field->skip)
+            if (!field->skip) {
```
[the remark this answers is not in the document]
emturner: This is different to the `skip button` - this is about whether a particular field should ever be shown to a user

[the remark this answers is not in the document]
ajinkyaraj-23: Better to leave 'skip' related stuff for another MR. 
Can you explain why only operation_step_operation has `state->is_field_complex = ..` what about other switch cases?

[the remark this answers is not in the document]
emturner: IIUC, this is for the point where we start parsing a new field?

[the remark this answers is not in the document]
spalmer25: Yes, the `TZ_OPERATION_STEP_OPERATION` is the step where we start the parsing of each field and decide how to handle them. The other step doesn't know much about the field itself.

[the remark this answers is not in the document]
ajinkyaraj-23: Great.

tests/unit/ctest/tests_parser.c
```diff
@@ -0,0 +1,398 @@
+
+#include <stdlib.h>
+#include "ctest.h"
+#include "../../../app/src/parser/operation_parser.h"
```
emturner, line 4: could consider `#include "../parser/operation_parser.h"` instead?

emturner: since it's copied there from the `app/src/parser` dir

tests/unit/ctest/tests_parser.c
```diff
@@ -0,0 +1,398 @@
+
```
emturner, line 1: could you add a copyright header?

tests/unit/ctest/tests_parser.c
```diff
@@ -0,0 +1,398 @@
...
+        default:
+            CTEST_ERR("%s:%d parsing error: %s", __FILE__, __LINE__,
+                      tz_parser_result_name(st->errno));
+        }
+        break;
+    }
+}
+
+CTEST2(operation_parser, check_proposals_complexity)
```
emturner, line 135: cheeky request - it would be awesome, if the `fields_check` had the parsed representation of the fields available too/were checked?

That way could add in assertions that e.g. a field of `Unit` doesn't actually get displayed etc?

spalmer25: I'm not sure I understand.
Would it be possible to address this change in another issue?

tests/unit/ctest/tests_parser.c
spalmer25: I'm not really into using file copies from other tests.
Would a solution like a symbolic link to the parser folder (and therefore having a `#include "parser/operation_parser.h"`) could work?

ajinkyaraj-23: You can simply use "operation_parser.h" and add include paths in compilation command in makefile.

emturner: makes sense! I like ajinkya's suggestion too

tests/unit/ctest/tests_parser.c
emturner: Yeah, makes sense 👍🏻 

As in, if we get a field `Code: Unit`, then we a) don't want to display that at all, and b) shouldn't set the `complex` field tag?

emturner (approved):

ajinkyaraj-23 (approved):
LGTM

PR #160, opened 2023-12-07
Boostrap the `ctest` framework for future unit tests written in C

emturner:
Lifted from: https://github.com/bvdberg/ctest 
Luckily, integration is easy as it is also Apache 2.0 licenced!


ajinkyaraj-23 (commented):
Can you add one sample test on how to use this ctest in given context?
How ctest will link with parser c files? where the test_x.c should be added?

Makefile
```diff
@@ -73,12 +73,14 @@ clean:
...
-		test/samples/operations/nano/samples.hex\
+		test/samples/operations/nano/samples.hex	\
 		tests/unit/parser/*.ml				\
 		tests/unit/parser/*.[ch]			\
 		tests/unit/parser/dune				\
 		tests/unit/parser/Makefile			\
-		tests/unit/Makefile
+		tests/unit/Makefile				\
+		tests/unit/ctest/Makefile			\
+		tests/unit/ctest/*.[ch]
```
spalmer25, line 83: Would it make sens / be possible to have those in the `tests/unit/Makefile` ?

spalmer25 (commented):
The `.gitignore` in `tests/unit/` could also be moved to `tests/unit/parser/`.

ajinkyaraj-23 (approved):
LGTM

Makefile
emturner: I'm not completely sure why this was needed either - could we merge as is and then modify in future?

spalmer25 (approved):

PR #169, opened 2023-12-14
Swap: add a doc and a script to perform swap tests.

spalmer25:
This PR add a script to ease swap tests snapshots update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,3 +223,45 @@ open at the end of a test.
 You will be requested to press enter to take each snapshot in term.
 **NB** make sure that the screen has updated to the screen you want to snapshot each time. It's also a good idea to
 re-run the test normally afterwards, to ensure the snapshots have been set correctly.
+
+## Swap test
+
+### Requirement
+
+Our swap tests are located in the https://github.com/functori/app-exchange repository.
+You must have a clone of this repository. In the commands below, it must be referenced in the `APP_EXCHANGE_REPO` variable.
+
+In some of the commands below, other variables will be required:
+- `APP_TEZOS_REPO`: your clone of the https://github.com/trilitech/ledger-app-tezos-wallet repository
+- `APP_ETH_REPO`: your clone of the https://github.com/LedgerHQ/app-ethereum repository
+The commands below will consider the version of the current branch of each repository.
+
+### Preparation
+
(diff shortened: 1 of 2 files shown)


scripts/test_swap.sh
```diff
@@ -0,0 +1,197 @@
...
+
+    device=$1
+    name=$2
+    repo=$3
+    name_elf=$name"_"$(echo $device | tr 'sp' 's2')".elf"
+    params="COIN=$name CHAIN=$name DEBUG=1"
+
+    _build_app $device $repo $params
+
+    sudo cp bin/app.elf $APP_EXCHANGE_REPO/test/python/lib_binaries/$name_elf
```
emturner, line 131: why is `sudo` required?

README.md
```diff
@@ -223,3 +223,40 @@ open at the end of a test.
...
+./scripts/test_swap.sh build_app_exchange $DEVICE
+./scripts/test_swap.sh build_app_ethereum $DEVICE
+./scripts/test_swap.sh build_app_tezos    $DEVICE
+```
+
+### Running
+
+You can run an individual test with :
+```sh
+./scripts/test_swap.sh run_tests --device $DEVICE -v -k "your_test_name"
```
emturner, line 253: could you add how to run all? without `-k "your test name"`?

emturner: oh, with `-k "tezos"`?

emturner: would it be possible to just have a `run_tests_all` for all the ones we care about? (like `update`)

README.md
```diff
@@ -223,3 +223,40 @@ open at the end of a test.
...
+```
+
+### Update
+
+You can run :
+```sh
+./scripts/test_swap.sh update $DEVICE
```
emturner, line 260: `update` is not picking up `APP_ETH_REPO` for me, but `run_tests` was

emturner: I think when it tries to run `build_app_ethereum`

scripts/test_swap.sh
spalmer25: I'm not sure why, probably because the copy file was created from docker. 
I'll try to get rid of it.

README.md
spalmer25: Fixed: it was a relative/absolute path issue.

emturner (approved):

PR #170, opened 2023-12-15
Parser: fix is_field_complex

spalmer25:
Since `field_name` are not operation fields, they do not need to be marked as complex.

This MR resets the `is_field_complex` field when there are no more fields to parse for an operation.

### DIFF
--- a/app/src/parser/operation_parser.c
+++ b/app/src/parser/operation_parser.c
@@ -703,6 +703,7 @@ tz_operation_parser_step(tz_parser_state *state)
             tz_stop(IM_FULL);
 
         if (name == NULL) {
+            state->is_field_complex = false;
             tz_must(pop_frame(state));
         } else {
             uint8_t present = 1;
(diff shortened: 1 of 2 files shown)


app/src/parser/operation_parser.c
```diff
@@ -703,6 +703,7 @@ tz_operation_parser_step(tz_parser_state *state)
             tz_stop(IM_FULL);
 
         if (name == NULL) {
+            state->is_field_complex = false;
```
ajinkyaraj-23, line 706: Two questions: 
1. What happens when a field in the middle is complex and then there is a field with name NULL? is that possible?
2. state->is_field_complex - is it an OR operation? because for each field we are reassigning it at line 714. So one field is compelex and next field is not complex then state->is_field_complex will be false at the end of parsing. Parser_state is shared among the fields right?

spalmer25: 1. `name == NULL` is used to detect that there are no more fields to parse for the operation, so only the last field is followed by `name == NULL`.
2. Yes, `state->is_field_complex` is shared between fields. It corresponds to the complexity of the current field.
At line 174, since we've changed fields, we redefine `state->is_field_complex` with the `complex` value of the new field, so effectively if a complex field is followed by a non-complex field, then `state->is_field_complex` will change from `true` to `false`.

ajinkyaraj-23 (approved):
LGTM

PR #173, opened 2023-12-19
Expert mode flow on nano

### LINKED ISSUE #167: Add expert mode for nano devices
- Refer to following workflow image for expert mode operation
 - use the expert mode boolean in global settings to check whether expert mode is enabled or not. 
 
 
image

spalmer25:
Expert mode is not used for now.

This PR adds a splash screen before the complex fields, as indicated by the specification.

fixes #167


tests/integration/nano/test_sign_simple_transaction.py
```diff
@@ -0,0 +1,46 @@
...
+from utils.app import nano_app, Screen, DEFAULT_ACCOUNT
+from utils.message import Message
+
+# Operation (0): Transaction
+# Fee: 0.5 XTZ
+# Storage limit: 4
+# Amount: 0.01 XTZ
+# Destination: KT18amZmM5W7qDWVt2pH6uj7sCEd3kbzLrHT
+
+if __name__ == "__main__":
```
emturner, line 27: I think sending it to a `entrypoint` should be fine, as long as parameter is still `unit`?

tests/integration/nano/nanox/regression_batched_ops.py
```diff
@@ -43,7 +43,7 @@
     test_name = Path(__file__).stem
     with nano_app() as app:
 
-        app.assert_screen(Screen.Home)
+        app.setup_expert_mode()
```
emturner, line 46: it looks like `entrypoint` gets a splash screen, but then `parameter` doesn't?

spalmer25: Yes, this is because the transaction `parameter` field is the representation of the OCaml `(entrypoint, parameter) option` type.
If this is possible, I'd like to find a solution in another issue.

tests/integration/nano/test_sign_simple_transaction.py
spalmer25: Yes, indeed.
Since this is a consequence of the problem raised in the comment below, I will correct this test when I deal with the problem.

app/src/apdu_sign.c
```diff
@@ -160,11 +160,39 @@ refill_blo_im_full(void)
...
+    if (st->field_info.is_field_complex && !N_settings.expert_mode) {
+        tz_ui_stream_push(TZ_UI_STREAM_CB_NOCB, st->field_info.field_name,
+                          "Needs Expert mode", TZ_UI_LAYOUT_HOME_BP,
+                          TZ_UI_ICON_NONE);
+        tz_ui_stream_push(TZ_UI_STREAM_CB_REJECT, "Home", "",
+                          TZ_UI_LAYOUT_HOME_PB, TZ_UI_ICON_BACK);
+        tz_ui_stream_close();
+    } else {
+        if (st->field_info.is_field_complex
+            && global.keys.apdu.sign.u.clear.last_field_index
```
ajinkyaraj-23, line 174: Where do you reset/initialize this global variable? Especially at the beginning and end of transaction?

ajinkyaraj-23 (commented):
LGTM. Where do you rese the clear.last_index?

app/src/apdu_sign.c
spalmer25: I do not.
Let me fix it.

ajinkyaraj-23: Ok. Looks good.

tests/integration/nano/nanox/regression_batched_ops.py
emturner: I see, thanks!
Ideally, would be able to find a work around, I feel it's possibly a bit confusing

emturner (approved):

PR #183, opened 2024-01-04
Add reject confirm screen for blindsigning flow

### LINKED ISSUE #181: Add confirm reject screen to blindsigning
Currently  in the blindsigning flow, clicking reject transaction immediately rejects the transaction. There should be confirm rejection screen in between just like it is for expert mode.

ajinkyaraj-23:
fixes #181


app/src/ui_stream.h
```diff
@@ -177,3 +177,4 @@ void   tz_ui_stream_close(void);
 void   tz_ui_stream(void);
 void   tz_ui_stream_start(void);
 tz_ui_cb_type_t tz_ui_stream_get_cb_type(void);
+void            pass_from_clear_to_blind(void);
```
emturner, line 180:
```suggestion
```

Don't need to expose it here?

app/src/ui_stream_nbgl.c
```diff
@@ -28,8 +28,9 @@ bool tz_ui_nav_cb(uint8_t, nbgl_pageContent_t *);
 bool has_final_screen(void);
 void tz_ui_stream_start(void);
 
-void drop_last_screen(void);
-void push_str(const char *, size_t, char **);
+void        drop_last_screen(void);
+void        push_str(const char *, size_t, char **);
+extern void pass_from_clear_to_blind(void);
```
emturner, line 33: this seems a bit too invasive, you should instead be calling a calling `s->cb(TZ_UI_STREAM_CB_BLINDSIGN)` here?

(you might need a *small* wrapper function in the file todo so)

app/src/ui_stream.h
ajinkyaraj-23: We have moved the blindsigning code to ui_stream_nbgl. There it calls to pass_from_clear_to_bind when blindsigning is approved. pass_from_clear_to_bind is in apdu_sign.c. So to access it needs to be exposed in a header file exposed to ui_stream_nbgl.c. Which is ui_stream.h.

app/src/apdu_sign.h
```diff
@@ -69,4 +69,5 @@ typedef struct {
 
 #ifdef HAVE_NBGL
 void switch_to_blindsigning(const char *, const char *);
```
emturner, line 71:
```suggestion
```
switch to blindsigning is no longer defined in `apdu_sign.c`, should be deleted.

ajinkyaraj-23 (commented):
> this seems a bit too invasive, you should instead be calling a calling `s->cb(TZ_UI_STREAM_CB_BLINDSIGN)` here?

(you might need a *small* wrapper function in the file todo so) 

Sorry this extern is redundant actually. Removed it.

app/src/apdu_sign.h
ajinkyaraj-23: yes.

app/src/ui_stream_nbgl.c
ajinkyaraj-23: Sorry this extern is redundant actually. Removed it.

app/src/apdu_sign.h
```diff
@@ -69,4 +69,5 @@ typedef struct {
...
+void tz_reject_ui(void);
```
emturner, line 72:
```suggestion
```
This isn't defined in `apdu_sign.h` - instead you can add it to `ui_stream.h`

app/src/ui_stream.h
emturner: Exactly, the only place that calls this is `ui_stream_nbgl`?

So you don't need to expose it in a header file at all. It's an implementation detail then. You can forward declare it at the top of `ui_stream_nbgl.c`?

ajinkyaraj-23: This function is shared with nano implementation as well. And it is called in apdu_sign.c:358 also.

app/src/ui_stream_nbgl.c
emturner: It shouldn't be forward declared here at all - you can (if necessary) add it to the `apdu_stream.h` file

But even that's not needed - instead of calling this function directly, you should call the stream callback?

app/src/ui_stream.h
emturner: Yes? Exactly - it's defined and called in `apdu_sign.c` first and foremost? So if you have to declare it in a header, `apdu_sign.c` is the correct header?

ajinkyaraj-23: Got it.

app/src/ui_stream_nbgl.c
ajinkyaraj-23: Will implement the Wrapper function .

app/src/apdu_sign.h
ajinkyaraj-23: Moving to ui_stream.h

emturner (approved):

PR #189, opened 2024-01-08
Swap: allow one reveal before the transaction

### LINKED ISSUE #188: Swap: allowing to sign a reveal followed by a transaction.
# Currently

During a swap, it is only permitted to sign batches of operations which contain only one operation and if this operation is one transaction.

# Goal

During a swap, to also be able to sign batches of operations that only contain one reveal followed by one transaction.

spalmer25:
fixes #188


emturner (approved):

ajinkyaraj-23 (approved):
Kudos to Sebastian for this quick fix.

PR #185, opened 2024-01-05
CI: use default app-exchange repo

emturner:
<test> - can we do away with the app-exchange fork?

PR #191, opened 2024-01-08
Dont update display while processing transactions called from swap.

### LINKED ISSUE #186: Swap: do not display anything when swapping
# Currently

When a swap is performed, the information about the operation we want to sign is displayed.
The Exchange application that is called before our application already displays this information.

# Goal

No UI if app_main is called from swap mode

ajinkyaraj-23:
fixes #186


emturner:
testing if swap fixed in #185

PR #198, opened 2024-01-18
Swap: Do not use bss memory in tezos swap library functions

spalmer25:
Adding MACRO in `exception.h` in order to return the error instead of putting it in the global variable `global`


app/src/handle_swap.c
```diff
@@ -46,6 +46,7 @@ void
 swap_handle_check_address(check_address_parameters_t *params)
 {
     FUNC_ENTER(("params=%p", params));
```
emturner, line 48: could switch to `PREAMBLE` at the same time?

app/src/handle_swap.c
```diff
@@ -46,6 +46,7 @@ void
...
+    tz_exc _sw_ret_code = 0x0000;
 
     if (params->address_to_check == NULL) {
         PRINTF("[ERROR] Address to check is null\n");
```
ajinkyaraj-23, line 52: change error to bail so you need only one flag.

emturner (approved):
lgtm

ajinkyaraj-23 (approved):

PR #203, opened 2024-02-09
Address audit report comments

### LINKED ISSUE #202: Fix Audit security comments
in app/src/apdu_sign.c : 486, buffer’s content as to be zeroized (i.e.,cleared). For this, only the buffer size was reset to zero. In doing so, the data persists in memory, with the risk of being read by another channel.

- [x] Reset the buffer and not just its size. 
- [x] Initialize `pressed_right` in ui_stream.c and ui_stream_nbgl.c
- [x] Fix TODO/NOTE and other comments - check app/src/ui_stream_nbgl.c: 352-354 and app/src/globals.h: 122
- [x] Remove redundant check for global.step in app/src/apdu_sign.c : 486
- [x] <=0 check is not correct for unsigned int. Fix at app/src/parser/operation_parser.c: 418
- [x] fix printf format specifier incompatibility at app/src/apdu_sign.c : 585

ajinkyaraj-23:
fixes #202

### DIFF
--- a/app/src/apdu.c
+++ b/app/src/apdu.c
@@ -1,7 +1,5 @@
 /* Tezos Ledger application - Some common primitives and some command handlers
 
-   TODO: split this file (apdu primitives and apdu handlers)
-
    Copyright 2023 Nomadic Labs <contact@nomadic-labs.com>
    Copyright 2023 Trilitech <contact@trili.tech>
 
--- a/app/src/apdu.h
+++ b/app/src/apdu.h
@@ -1,7 +1,5 @@
 /* Tezos Ledger application - Some common primitives and some command handlers
 
-   TODO: split this file (apdu primitives and apdu handlers)
-
    Copyright 2023 Nomadic Labs <contact@nomadic-labs.com>
 
    With code excerpts from:
--- a/app/src/apdu_sign.c
+++ b/app/src/apdu_sign.c
@@ -488,15 +488,14 @@ handle_first_apdu_clear(__attribute__((unused)) command_t *cmd)
 #endif
         tz_ui_stream_init(stream_cb);
         global.step = ST_CLEAR_SIGN;
-        if (global.step == ST_CLEAR_SIGN) {
+
 #ifdef TARGET_NANOS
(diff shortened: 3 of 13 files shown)


app/src/apdu.c
```diff
@@ -1,7 +1,5 @@
 /* Tezos Ledger application - Some common primitives and some command handlers
 
-   TODO: split this file (apdu primitives and apdu handlers)
```
spalmer25, line 3: I'm not sure whether we have any trace of these `TODO`s.
Could you open an issue to make sure we don't forget them?

spalmer25 (commented):
LGTM

app/src/apdu.c
ajinkyaraj-23: Created issues #207 #208 and #209

ajinkyaraj-23:
Rebased on main.

app/src/apdu.c
spalmer25: What about ?
```
        // TODO: this is handled by change_screen_right except we disable it
        // to use it here. We should make ui_stream fully compatible with
        // exception.h
```
in [app/src/ui_stream_nbgl.c](https://github.com/trilitech/ledger-app-tezos-wallet/pull/203/files#diff-ba33a3ab1ba1a41b1110b6ddf932fc7096b6f05730525c30fb5ffc398fc6dcce)
Is this `TODO` outdated ?

spalmer25 (commented):
There is a forgotten `FIXME` in https://github.com/trilitech/ledger-app-tezos-wallet/blob/5b4e04c247bd801c9d78251883e3167051d4922c/app/src/ui_stream.h#L154-L155

app/src/apdu.c
ajinkyaraj-23: Created #211 .

ajinkyaraj-23:
> There is a forgotten `FIXME` in
> 
> https://github.com/trilitech/ledger-app-tezos-wallet/blob/5b4e04c247bd801c9d78251883e3167051d4922c/app/src/ui_stream.h#L154-L155

Fixed. The issue is same as #211

spalmer25 (approved):

PR #204, opened 2024-02-12
Tests: fix failing tests at app-builder version 3.15.0

spalmer25:
- [x] Solution: set the `app-builder` version to 3.14.0


ajinkyaraj-23 (approved):
LGTM

PR #246, opened 2024-07-09
Use the guideline provided useCaseHomeAndSettings use case.

ajinkyaraj-23:


tests/integration/stax/utils.py
```diff
@@ -110,8 +110,12 @@ def make_golden(self):
         for filename in os.listdir(path):
             os.remove(os.path.join(path, filename))
         path = f"{self.__snapshots_path}"
-        os.remove(os.path.join(path, "home.png"))
-        os.remove(os.path.join(path, "info.png"))
+        try:
+            os.remove(os.path.join(path, "home.png"))
+            os.remove(os.path.join(path, "info.png"))
+        except FileNotFoundError:
+            pass
```
spalmer25, line 117: We could imagine that the image `info.png` would exist, but not `home.png`. In this case, `info.png` will not be removed 
```suggestion
        home_path=os.path.join(path, "home.png")
        info_path=os.path.join(path, "info.png")
        if os.path.exists(home_path): os.remove(home_path)
        if os.path.exists(info_path): os.remove(info_path)
```

spalmer25 (approved):

PR #249, opened 2024-07-16
Stax tests update

spalmer25:
`Stax` display has been updated.

This PR updates `Stax` tests accordingly.

Two of the display changes will require changes to our code:
 - A `Warning` button has been added to some screens. We should understand when we want them to be displayed and have a test to ensure its behaviour: https://github.com/trilitech/ledger-app-tezos-wallet/issues/250.
image
 It is also not displayed well in some cases: https://github.com/trilitech/ledger-app-tezos-wallet/issues/251. For example:
image
 - A number of pages has been added to some flow and seems to not correspond to the right number of pages. https://github.com/trilitech/ledger-app-tezos-wallet/issues/252
image


ajinkyaraj-23 (approved):
LGTM

PR #255, opened 2024-07-18
Add michelson `D_Ticket` to the formatter

spalmer25:


app/src/parser/formatting.h
```diff
@@ -23,7 +23,7 @@
  *
  *        Should be kept in sync with the last protocol update.
  */
-#define TZ_LAST_MICHELSON_OPCODE 157
+#define TZ_LAST_MICHELSON_OPCODE 158
```
ajinkyaraj-23, line 27: Last opcode is from 0 index or 1 index. I see below that last index is 157.
You are already using last_michelson_op_code +1 in the array definition

ajinkyaraj-23 (changes requested):

app/src/parser/formatting.h
spalmer25: That's right !
Fixed

PR #262, opened 2024-07-24
261 duplicate rejection pages at last signing page

ajinkyaraj-23:
The issue occurs because usecaseReviwStreamingFinish creates a rejection confirmation screen on its own. This no need to add it in the code.


spalmer25 (approved):
Could you explain in the issue or in the PR's description why a such behaviour occurs?
Otherwise, it looks good to me.

PR #269, opened 2024-08-22
Add a summary of transaction parsed for batch transactions. 

ajinkyaraj-23:


app/src/parser/num_parser.c
```diff
@@ -97,3 +97,63 @@ tz_string_to_mutez(const char *str, uint64_t *res)
...
+        return false;
+    }
+
+    uint64_t integer_part = microtez / 1000000;
+    uint64_t decimal_part = microtez % 1000000;
+
+    // Convert integer part to string
+    size_t i = 0;
+    if (integer_part == 0) {
+        str[i++] = '0';
```
spalmer25, line 114: Checks are required to prevent Out-Of-Bound issue

tests/integration/nano/snapshots/nanos/test_nanos_regression_batched_ops/00035.png
spalmer25, line 1: Where can I find the display specification? If there isn't one, we need to discuss it and create a detailed one.

app/src/apdu_sign.c
```diff
@@ -247,7 +272,9 @@ refill_blo_done(void)
         TZ_CHECK(sign_packet());
         TZ_SUCCEED();
     }
-
+    if (op->batch_index > 1) {
+        tz_ui_stream_push_summary();
+    }
```
spalmer25, line 277: When do the user have to see the summary?
From what I understand, it is displayed after all the fields has been display.
Is it the expected behaviour?

app/src/parser/operation_parser.c
```diff
@@ -662,22 +662,20 @@ tz_step_read_num(tz_parser_state *state)
...
             tz_raise(INVALID_DATA);
         }
         switch (op->frame->step_read_num.kind) {
         case TZ_OPERATION_FIELD_AMOUNT:
-            op->last_amount = value;
+            op->total_amount += value;
             break;
         case TZ_OPERATION_FIELD_FEE:
-            op->last_fee = value;
+            op->total_fee += value;
```
spalmer25, line 674: Adding fees will change the behaviour of `SWAP`:

`SWAP` authorized 2 kind of transaction:
 - only one `Transaction` operation.
   - `Adding` or `Taking the last` amount (respectively, fee) leads obviously to the same result.
 - one `Reveal` followed by one `Transaction`.
   - Since there is no amount in `Reveal` operation, `Adding` or `Taking the last` amount leads to the same result.
   - But there are fees in `Reveal` and `Transaction`, so `Adding` or `Taking the last` fee will change what will be displayed.

We need to ask Ledger what should be the fees displayed in the second case.

The [Ledger's SWAP tests](https://github.com/LedgerHQ/app-exchange/blob/develop/test/python/test_tezos.py#L31) do not handle the second case. Maybe we should think about adding a test for it.

ajinkyaraj-23: Probably this was causing the bugs for new accounts on tezos. Can we create a swap test for this on their repo?

tests/integration/nano/test_sign_batch_operation.py
```diff
@@ -0,0 +1,45 @@
...
+
+from pathlib import Path
+
+from utils.app import Screen, DEFAULT_ACCOUNT
+from utils.message import Message
+
+# Operation (0): Reveal
+# Fee: 0.01 XTZ
+# Storage limit: 4
+# Public key: edpkuXX2VdkdXzkN11oLCb8Aurdo1BTAtQiK8ZY9UPj2YMt3AHEpcY
```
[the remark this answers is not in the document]
ajinkyaraj-23: fixed.

ajinkyaraj-23:
Merged in #283

Issue #268, opened 2024-08-21
Create a blindsigning settings


ajinkyaraj-23:
Blindsigning allowed with three possible options
- BLindsign  for large Tx
- BLindsigning ON
- Blindsigning OFF

ajinkyaraj-23:
Completed.

PR #271, opened 2024-08-23
Make nano home page according to Ledger app boilerplate

ajinkyaraj-23:
- [x] Change nano home and settings according to app boilerplate


ajinkyaraj-23:
UX update for stax is remaining. This PR only adds the settings. Does not use the settings. Need to create another PR where based on the settings the warnings and summary is displayed.

tests/integration/nano/snapshots/nanos/settings_max_screen_30.png
spalmer25, line 1: Why these values? Why 30 is the default value? Do we have a spec that requests this?
Is this the standard way to display this kind of setting, or do we have a spec that requests to display the max screen like this?

Another way to display this setting is to:
 - display the value of the `Max screens`:
 "Max screens"
  "30"
 - on BOTH click open a new flow that display a list of screens each displaying one possible value.
 BOTH click on one of the screen will set the indicated value and back to the "Max screens" screen in setting.
 The last screen is a back button that allow to come back to the setting page without changing the value.

There is probably many other way to do this. If there is no spec, let's discuss it and create the spec.

spalmer25 (commented):
Using the same template as the app boilerplate to display home and settings is not related to the issue #268. The new way of displaying the menu differs significantly from what was done and what is still used to display the rest of our app. We need to make sure that the behaviour of our app is unchanged. Let's open a second PR for this change.

app/glyphs/tezos_16px.gif
spalmer25, line 1: Using the same template as the app boilerplate to display home and settings is not related to the issue https://github.com/trilitech/ledger-app-tezos-wallet/issues/268. The new way of displaying the menu differs significantly from what was done and what is still used to display the rest of our app. We need to make sure that the behaviour of our app is unchanged. Let's open a second PR for this change.

tests/integration/nano/snapshots/nanos/settings_max_screen_30.png
spalmer25, line 1: Why these values? Why 30 is the default value? Do we have a spec that requests this?
Is this the standard way to display this kind of setting, or do we have a spec that requests to display the max screen like this?

Another way to display this setting is to:
 - display the value of the `Max screens`:
 "Max screens"
  "30"
 - on BOTH click open a new flow that display a list of screens each displaying one possible value.
 BOTH click on one of the screen will set the indicated value and back to the "Max screens" screen in setting.
 The last screen is a back button that allow to come back to the setting page without changing the value.

There is probably many other way to do this. If there is no spec, let's discuss it and create the spec.

app/src/ui_home.c
```diff
@@ -22,53 +22,46 @@
...
-    }
 
-    FUNC_LEAVE();
-}
 
-#endif  // HAVE_BAGL
+UX_STEP_NOCB(ux_menu_ready_step, pnn,{&C_tezos_16px , "Application", "is ready"});
+UX_STEP_NOCB(ux_menu_version_step, bn, {"Version", APPVERSION});
+UX_STEP_CB(ux_menu_settings_step, pb, ui_settings_init(SETTINGS_HOME_PAGE), {&C_icon_coggle, "Settings"});
+UX_STEP_VALID(ux_menu_exit_step, pb, os_sched_exit(-1), {&C_icon_dashboard_x, "Quit"});
```
spalmer25, line 39: Use `app_exit` instead of `os_sched_exit(-1)`

app/src/ui_home.c
```diff
@@ -22,53 +22,46 @@
...
 
+// FLOW for the main menu:
+// #1 screen: ready
+// #2 screen: version of the app
+// #3 screen: about submenu
```
spalmer25, line 44: What about ?
```suggestion
// #3 screen: settings menu
```

app/src/ui_home.c
```diff
@@ -22,53 +22,46 @@
...
-                      TZ_UI_LAYOUT_HOME_BN, TZ_UI_ICON_NONE);
-    tz_ui_stream_push(SCREEN_SETTINGS, "Settings", "", TZ_UI_LAYOUT_HOME_PB,
-                      TZ_UI_ICON_SETTINGS);
-    tz_ui_stream_push(SCREEN_QUIT, "Quit?", "", TZ_UI_LAYOUT_HOME_PB,
-                      TZ_UI_ICON_DASHBOARD);
-    tz_ui_stream_close();
-    tz_ui_stream_start();
+    if (G_ux.stack_count == 0) {
+        ux_stack_push();
+    }
```
spalmer25, line 61: What is the purpose of this?
Could you add a comment?

app/src/ui_settings.c
```diff
@@ -21,44 +21,34 @@
...
+UX_FLOW(ux_expert_mode_flow,
+        &ux_expert_mode_step,
+        &ux_back_step,
+        FLOW_LOOP);
+
 void
 ui_settings_init(int16_t page)
 {
-    const char *exp_mode = "DISABLED";
+    strncpy(global.expert_mode_state,"DISABLED", sizeof(global.expert_mode_state));
```
spalmer25, line 44: In order to prevent a not necessary copy, would it be possible to move this copy in the `else` part of the next `if` statement?

app/src/ui_home.c
spalmer25, line 39: `UX_STEP_VALID` is tagged as deprecated by Ledger:
https://github.com/LedgerHQ/ledger-secure-sdk/blob/master/lib_ux/include/ux_flow_engine.h#L177-L178
Use `UX_STEP_CB` instead

tests/integration/nano/test_basic.py
```diff
@@ -32,21 +32,25 @@ def test_basic(app):
...
     app.assert_screen(Screen.Quit)
     app.backend.right_click()
-    app.assert_screen(Screen.Quit)
+    app.assert_screen(Screen.Home)
+    app.backend.right_click()
+    app.assert_screen(Screen.Version)
 
     # Check Settings menu operation
-    app.backend.left_click()
+    app.backend.right_click()
```
spalmer25, line 53: Could you add inputs to the basic test scenario to ensure that the parameter screens loop (left-clicking on the first screen takes you to the last screen, and right-clicking on the last screen takes you to the first screen)?

spalmer25:
There are warnings during compilation. It would be better if there were none.

app/src/ui_home.c
ajinkyaraj-23: Done.

app/src/ui_home.c
ajinkyaraj-23: Done.

app/src/ui_home.c
ajinkyaraj-23: Commented. Its to reserve a stack spot if there is none.

app/src/ui_settings.c
ajinkyaraj-23: Done.

app/src/ui_home.c
ajinkyaraj-23: Done.

tests/integration/nano/test_basic.py
ajinkyaraj-23: Done.

ajinkyaraj-23:
> There are warnings during compilation. It would be better if there were none.

Removed warnings. Thanks

spalmer25 (approved):

PR #278, opened 2024-09-09
Add blindsigning page to stax

ajinkyaraj-23:


tests/integration/touch/test_basic.py
```diff
@@ -30,6 +30,12 @@
     app.settings.toggle_expert_mode()
     app.assert_settings()
 
+    app.settings.next()
+    app.assert_settings(blindsigning=True, blindsigning_status=0)
```
spalmer25, line 34: `blindsigning` seems always be `True`. Maybe you can just use `blindsigning_status` and remove the `blindsigning` argument.
Could you also name the Status ; It will be clearer than just [1,2,3]

app/src/globals.c
```diff
@@ -47,10 +47,17 @@ toggle_expert_mode(void)
...
+    tmp.blindsign_status = status;
     nvm_write((void *)&N_settings, (void *)&tmp, sizeof(N_settings));
 }
+
+void
+toggle_blindsign_status(void)
+{
+    blindsign_state_t status = (N_settings.blindsign_status + 1) % 3;
+    set_blindsign_status(status);
+}
```
spalmer25, line 63: Unused function. Can be used in the function `blindsign_toggle` of the file `ui_settings.c`  instead of `set_blindsign_status((N_settings.blindsign_status + 1) % 3);`

app/src/ui_home_nbgl.c
```diff
@@ -28,79 +28,116 @@
...
 {
+    FUNC_ENTER(("Token : %d, Index:  %d, Page: %d", token, index, page));
     uint8_t switch_value;
     if (token == EXPERT_MODE_TOKEN) {
         switch_value = !N_settings.expert_mode;
         toggle_expert_mode();
-        switches[EXPERT_MODE_TOKEN_ID].initState
-            = (nbgl_state_t)(switch_value);
+        expert_mode_switch.initState = (nbgl_state_t)(switch_value);
+        tz_ui_home_redisplay(EXPERT_MODE_PAGE);
```
spalmer25, line 98: I see here that the settings are redisplayed each time the `expert_mode toggle` or the `blindsign status` is changed.
Would it be possible to not redisplay the settings? And, probably, let the framework change it dynamically for us?

app/src/globals.c
ajinkyaraj-23: Done.

tests/integration/touch/utils.py
```diff
@@ -136,6 +136,11 @@ def toggle_expert_mode(self):
         """Toggle the expert_mode switch."""
         self._toggle_list.choose(1)
 
+    def set_blindigning(self, value: int):
+        if value not in [1, 2, 3]:
+            raise ValueError("Value must be 1, 2 or 3")
+        self._toggle_list.choose(value)
+
     def toggle_blindsigning(self):
```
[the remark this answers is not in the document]
ajinkyaraj-23: Done.

tests/integration/touch/test_basic.py
ajinkyaraj-23: Done.

app/src/ui_home_nbgl.c
ajinkyaraj-23: Removed from expert_mode_switch but cannot remove from blindsigning. When this refresh is removed, the selected radio choice resets to default one on page change.

tests/integration/touch/utils.py
```diff
@@ -136,14 +137,20 @@ def toggle_expert_mode(self):
...
+    def set_blindigning(self, value: int):
+        if value not in [1, 2, 3]:
+            raise ValueError("Value must be 1, 2 or 3")
+        self._toggle_list.choose(value)
 
     def exit(self) -> None:
         """Exits settings."""
         self.multi_page_exit()
 
+class blindsigning_status(Enum):
```
spalmer25, line 149: Could you use `CamelCase` for class name? (i.e. `BlindSigningStatus`)

tests/integration/touch/utils.py
```diff
@@ -136,14 +137,20 @@ def toggle_expert_mode(self):
...
+    NONE = 0
```
spalmer25, line 150: The `NONE` case does not have sense here. The `blindsigning_status` must be one of the 3 other options.

tests/integration/touch/utils.py
```diff
@@ -245,15 +252,12 @@ def assert_info(self):
         self.assert_screen("info", True)
 
     def assert_settings(self,
-                        blindsigning = False,
-                        expert_mode = False):
+                        expert_mode = False, blindsignStatus = blindsigning_status.NONE):
```
spalmer25, line 255: What would happen if the both parameter are not set (`self.assert_settings()`) or both set (`self.assert_settings(expert_mode=True, blindsigning_status=ON)`) ?
This `assert_settings` was initially created to be able to check the settings screen with multiple toggle.
Since the pages are different now, creating two functions make more sense here:
 - `assert_expert_mode(expert_mode : bool)` 
 - `assert_blindsigning_status(blindsigning_status : BlindSigningStatus)`

spalmer25 (approved):

PR #281, opened 2024-09-12
Reduce the global struct size

spalmer25:
At many times, when some fields were added on some structures, two tests were failing on **NanoS**:
 - `test_tz3_sign_micheline_basic`
 - `test_tz2_sign_micheline_basic.`
This is caused by an already known issue that signing with _tz2_ and _tz3_ needs more memory than for _tz1_.

This PR reduce the memory space required for the `global` by reordering structures fields.

Gain of `32*sizeof(uint8_t)` memory space.

### DIFF
--- a/app/src/apdu_sign.h
+++ b/app/src/apdu_sign.h
@@ -74,8 +74,8 @@ typedef struct {
     union {
         /// @brief clear signing state info.
         struct {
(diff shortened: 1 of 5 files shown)


ajinkyaraj-23 (approved):
LGTM

PR #284, opened 2024-09-12
Maximum one expert-mode warning per signing flow

### LINKED ISSUE #282: Maximum one expert-mode warning per signing flow
Display expert mode warnings only once for the entire (batch of) transaction. This is as per the google doc design we discussed with Ledger.

spalmer25:
fixes #282 

Currently, when a batch operation contains multiple fields requiring expert mode, the ledger displays a warning screen before each of them.

This PR adds changes to display the expert mode warning screen only before the first field that requires it.


ajinkyaraj-23 (approved):
LGTM

PR #289, opened 2024-09-19
[summary] remove too long operation warning on parsing error

### LINKED ISSUE #288: [blindsign]
# Currently
On nano devices, when the _blindsigning_ option is `Large Tx`, review a too long transaction:
 - display the first 20 screens of the operations
 - then display warning screens to warn that the operation is too long.
 - If the parser then fails to parse the rest of the operation, it displays warning screens to warn that the operation can not be parsed. 
 - On continue, it displays the hash only.
 
# Issue
Display two warning screens (One for the **Too long operation** and one for the **Parsing error**) is not user-friendly.
In this case the **Too long operation** is not relevant because the summary won't be displayed.
 
# Goal
If during a review a **Too long operation** is encountered:
If we do encounter a parsing error:
 - Display the parsing error warning flow
 - On accept switch too blindsign (display the hash)
If we do not encounter error:
 - Display the too long operation warning flow
 - On accept switch to summary (display the summary)

spalmer25:
fixes #288 
This PR change the order of handlers:

At too long operation (that will not be parsed successfully),

Before:
 - Display _too long operation_ warning screens
 - Then finish parsing

After:
 - Finish parsing
 - Then display _too long operation_ warning screens


ajinkyaraj-23 (approved):
LGTM

PR #294, opened 2024-09-26
[blind] Show blind warning on manual blind

spalmer25:
Currently, on nano device, when the _blindsigning_ option is `ON`, during a review, if the user clicks on the screen that allows him to blindsign (the 2<sup>nd</sup> screen), the summary is directly displayed. But we expect to show some warning screen before.

This PR add these warning screens.


ajinkyaraj-23 (approved):
LGTM

PR #297, opened 2024-09-30
[nano] remove blind large tx option

spalmer25:
As followed a discussion with ledger:
This PR 
 - use the behaviour of `Large tx` blindsign option for the `ON` blindsigning option
 - remove the `Large tx` blindsign option


app/src/globals.h
```diff
@@ -63,9 +63,14 @@ typedef enum {
 } main_step_t;
 
 typedef enum {
+#ifdef HAVE_BAGL
```
ajinkyaraj-23, line 66: use bool. there are only two settings so enum does not make sense.

app/src/globals.h
```diff
@@ -76,11 +81,13 @@ typedef enum {
 #define NB_MAX_SCREEN_ALLOWED 8
 #endif
 
+#ifdef HAVE_NBGL
```
ajinkyaraj-23, line 84: Are you sure you dont need the reason, 
You have to show amount, fees, hash  in case of large_tx but only hash in case of parsing error. How do you differenciate the two cases.

tests/integration/nano/test_blindsign.py
```diff
@@ -1,238 +0,0 @@
-#!/usr/bin/env python3
```
ajinkyaraj-23, line 1: where is the test to verify blindsign on behaviour now? for example for screens greater 12/20 for nanox/nanosp, also what happens when parsing error occurs after this limit when blindsigning is on.

app/src/globals.h
spalmer25: The two flow are different, so no need to differentiate the case later:
 - On parsing error the screen are directly displayed then the flow displayed is the blindsigning flow.
 - On too-large-operation the flow is the summary flow (finish the parsing before display warning screen)

app/src/globals.h
spalmer25: :+1: I will change it when I will apply my commits in your PR

tests/integration/nano/test_blindsign.py
spalmer25: Since the blindsigning flow as been replaced by the too-large-tx flow, the tests regarding blindsigning are for parsing_error, see:
- [test_blindsign_reject.py](https://github.com/trilitech/ledger-app-tezos-wallet/blob/main/tests/integration/nano/test_blindsign_reject.py)
- [test_blindsign_too_deep.py](https://github.com/trilitech/ledger-app-tezos-wallet/blob/main/tests/integration/nano/test_blindsign_too_deep.py)
- [test_blindsign_too_large.py](https://github.com/trilitech/ledger-app-tezos-wallet/blob/main/tests/integration/nano/test_blindsign_too_large.py)
- [test_parsing_errors.py](https://github.com/trilitech/ledger-app-tezos-wallet/blob/main/tests/integration/nano/test_parsing_errors.py)
- [test_sign_too_long_operation.py](https://github.com/trilitech/ledger-app-tezos-wallet/blob/main/tests/integration/nano/test_sign_too_long_operation.py)

spalmer25:
The commits of this PR have been merged to the PR #296

PR #298, opened 2024-10-02
Fix static analysis warning for unreachable BAGL code for NBGL

ajinkyaraj-23:


spalmer25 (approved):
It could be great to have the static analysis on our project too

PR #301, opened 2024-10-31
[test] explicits messages

spalmer25:
depends on #300 
Raw messages in tests make them less clear and hard to modify.
This PR adds ways to build and encode messages to better understand and maintain tests.

PR #303, opened 2024-11-29
[test] sort and merge nano tests

spalmer25:
depends on #301 
This PR groups tests into folders and files to have a better global understanding of which tests exists.
Use `pytest` feature for better intelligibility.


ajinkyaraj-23 (approved):
LGTM

PR #308, opened 2025-01-21
Adapt Nano tests for Stax/Flex devices

spalmer25:
There are many tests for Nano devices and few for Stax/Flex devices.
The two types of test were separated because they were too different, but the Nano tests have been updated to a more recent version, so the two are now similar enough to be merged.
This MR adapts the Nano tests for Stax/Flex devices.


ajinkyaraj-23 (approved):
LGTM

PR #314, opened 2025-06-16
Bump tezos version to Seoul

spalmer25:
This PR, according to the changes made in the protocol Seoul:
 - adds change the encoding for the signature to ensure that the signature is a BLS signature. (because it is only used for BLS signature reading)
 - adds the `proof` field to the reveal operation
 - adds a `update_companion_key` operation


tests/integration/python/snapshots/nanosp/test_sign/operations/test_sign_reveal/test_operation_flow[flow-with_proof]/00007.png
Comment (ajinkyaraj-23) [inline tests/integration/python/snapshots/nanosp/test_sign/operations/test_sign_reveal/test_operation_flow[flow-with_proof]/00007.png:1]: why are we not showing the proof in this reveal for bls signature. Isn't the whole point of adding proof was to show proof of ownership?
Also are we checking if the proof is valid? is that our responsibility?

tests/integration/python/snapshots/flex/test_sign/operations/test_sign_set_companion_key/test_operation_flow[flow-basic]/00002.png
Comment (ajinkyaraj-23) [inline tests/integration/python/snapshots/flex/test_sign/operations/test_sign_set_companion_key/test_operation_flow[flow-basic]/00002.png:1]: again we are not showing the proof in the UI

Comment (spalmer25) [reply tests/integration/python/snapshots/flex/test_sign/operations/test_sign_set_companion_key/test_operation_flow[flow-basic]/00002.png:1]: The proof for `reveal`, `update_consensus_key` and `update_companion_key` have the same purpose and are all safe for the user.

ajinkyaraj-23:
But it goes against the idea of clear signing. At least in Reveal we need
to show the proof, as other wallets will be showing it too.
I would say show the proof, Reveal is a small operation anyways. Same goes
for update_consensus_keys and companion key. What the user signs and what
he sees should be the same. These are all one time operations. and proof is
shown only when tz4 BLS address is used.

Ajinkya Rajandekar

Embedded Software Engineer

LinkedIn <https://www.linkedin.com/in/ajinkyaraj/>

